PR #94, opened 2020-06-04
Update doc on configuring launch target (IEP-199)

Julien-Kinderf:
This step needed an extra eclipse plugin that are not here by default. I just clarified the thing.


kolipakakondal:
@Julien-Kinderf Eclipse CDT by default comes with the launch bar controls at least from the CDT 8.5 which was released in 2015 I think.

PR #155, opened 2020-11-11
IEP-274: Address Indexer and build configuration problems(IEP-276)

kolipakakondal:
- Raised plugins compiler compliance to Java 11
- Updated to new CompileCommandsJsonParser(cmake4eclipse parser)
- Implementing CBuildConfiguration directly
- CDT 10.0 and Eclipse 2020-09 build infra
- Other enhancements


kolipakakondal:
Hi All, I'll add the next set of changes which I'm currently working on in the different PR as we already created different tickets. That also makes our life easy for review, testing, and quick releases. This a base PR that addresses the CMake4Eclipse parser migration

**Test cases covered:**
1. Build - Are errors got resolved after the build? Yes
2. Once the errors got resolved - close the editor and reopen it - Are the errors coming back? No
3. Re-run the build - Is it building the entire project? No
4. Make some changes in the C files/CMakeLists.txt - Does it build correctly? Yes
5. Restart eclipse and see the errors appear again? No
6.  Are we able to build with target-specific esp32 and esp32s2? Yes

This requires Java 11 and Eclipse 4.17 With this, we won't support previous versions of Eclipse and there is no way to achieve that as well as CMake4Eclipse is added in Eclipse 4.17

sigmaaa (approved):
I checked the mentioned test cases on Windows 10. I used 3 different project templates: sample_project, hello_world, and blink. Everything works fine as expected.

kolipakakondal:
Hi @igrr - Can we merge this and release 1.3.0? Just thinking to push the releases progressively rather than waiting so long for everything to be fixed.

The only problem would be we have to force the users to update to eclipse 4.17 which is unavoidable in a way since CMake4Eclipse parser update is available only from Eclipse 4.17. Eclipse 4.18 is also released a few days back but we are not there yet to support it. We can plan this in the next release.

igrr:
@kolipakakondal re. eclipse 4.17 requirement — i think it is fine as long as it is prominently mentioned in the release notes. That is, if the users who have an older version of Eclipse get the new version of the plugin via auto-update feature, we should warn them somehow. Is there a way to specify compatible Eclipse version in the plugin manifest, to prevent installation on an unsupported version of Eclipse?

Also it might be a good idea to bump the version to 2.0.0 in this case — if we try to stick to SemVer, this is a "breaking change".

kolipakakondal:
>  Is there a way to specify compatible Eclipse version in the plugin manifest, to prevent installation on an unsupported version of Eclipse?

@igrr - Yes, this can be done by defining the feature/plugin dependency in the idf feature.xml. I'll check this.

For example: With CDT 10.0 (in Eclipse 4.17), `org.eclipse.cdt.cmake.is.core` is the new plugin which is added for CMake Parser udpate. We can add this plugin as the pre-requisite to install idf plugin.

https://github.com/espressif/idf-eclipse-plugin/blob/master/features/com.espressif.idf.feature/feature.xml#L103

```
  <requires>
      <import plugin="org.eclipse.cdt.cmake.is.core"/>
   </requires>
```

PR #163, opened 2020-12-08
Added search tree filter to templates (IEP-300)

sigmaaa:
added search filter based on PatternFilter, which is searching elements in the tree structure

IEP-294

### DIFF
--- a/bundles/com.espressif.idf.ui/src/com/espressif/idf/ui/templates/AbstractTemplatesSelectionPage.java
+++ b/bundles/com.espressif.idf.ui/src/com/espressif/idf/ui/templates/AbstractTemplatesSelectionPage.java
@@ -41,6 +41,7 @@ public abstract class AbstractTemplatesSelectionPage extends BaseWizardSelection
 {
 	protected TreeViewer templateViewer;
 	protected ITemplateNode templateElements;
+	protected FilteredTree filteredTree;
 	private WizardSelectedAction doubleClickAction = new WizardSelectedAction();
 
 	private class WizardSelectedAction extends Action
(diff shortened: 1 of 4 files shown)


kolipakakondal (changes requested):
1. Could you also enable the filter control when only the 'create a project using one of the templates` check box is checked? Otherwise, it will lead to confusion - users will type something and feel that it's not filtering correctly!
2. When you clear the filter text, the viewer should show the complete tree again
3.  It seems to be viewer is not refreshing after entering the filter text - I have to click one of the element to see the filtered content. 

I'm on macOS 10.15.7

sigmaaa:
Now filter's text field is enabled only when the checkbox is selected. When the user moves the cursor over a disabled item, a tooltip appears: 
IEP-294 disabled field

kolipakakondal:
> * When you clear the filter text, the viewer should show the complete tree again
> * It seems to be viewer is not refreshing after entering the filter text - I have to click one of the element to see the filtered content.

@sigmaaa Did you have a chance to verify the above scenarios?

sigmaaa:
> > * When you clear the filter text, the viewer should show the complete tree again
> > * It seems to be viewer is not refreshing after entering the filter text - I have to click one of the element to see the filtered content.
> 
> @sigmaaa Did you have a chance to verify the above scenarios?

I tried on MacOS v. 10.15.7 and it looks the same as on windows for me. The next gif is showing functionality. I'm not clicking on field to refresh it, it's refresh automatically and after I'm clearing the filter text, the viewer is showing complete tree.
IEP-298-demonstration

PR #171, opened 2020-12-22
Iep 53 imrpove filtering in sdkconfig editor

sigmaaa:
Now filtering based not only on the visible tree elements but on child elements also
IEP-53(2)
IEP-53


kolipakakondal (approved):
LGTM

PR #183, opened 2021-01-11
Optimized default wizard size

sigmaaa:
Optimized new project creation wizard size. More picture below:
Here on the left is a new size, and on the right is an old size.
IEP-331

Tested both on windows 10 and macOS 10.15.7.


kolipakakondal (approved):
LGTM. Please also check how this is working with the markdown text.

PR #185, opened 2021-01-12
IEP 308 list installed tools show an information dialog when there is an error

sigmaaa:
Now List Installed tools action shows information dialog to the user when it's clicked before installed the tools (when tools are not installed and IDF_PATH is not set). The information dialog looks like this:
IEP-308
After clicking "Yes", the Install tools dialog will be open. 
Test case:
- Go to Preferences -> C/C++ -> Build -> Environment and click Restore Defaults to clean environment variables.
-  Now go to Help -> ESP-IDF Tools Manager -> List Installed Tools and in the information dialog that appears click on "Yes". 
-  After this, the Install Tools dialog will be open, where it's necessary to choose ESP-IDF, git, and python locations. 
-  After tools installation environment variables are set again, and "List Installed Tools" is providing information about installed tools. 

Tested on Windows 10 and macOS 10.15.7


kolipakakondal (approved):
LGTM

PR #190, opened 2021-01-18
IEP-317 Fixing error jdk when not releasing sdkconfig.json after switching launch target

sigmaaa:
Now after switching the target from esp32 to esp32s2, the java process is releasing the lock on sdkconfig.json file so it can be deleted and not cause a build error. 
However, if a whole directory "build" is deleted after switching the target on Windows 10 ( I don't have the same issue on macOS),  there are other issue: "CreateProcess error=267, The directory name is invalid”, when I'm trying to build a project. 

To fix it, in deleteDirectory function was added condition statement, which leaves an empty build directory in the end. 

With both of these changes, the problem with building projects after switching targets is gone on both Windows 10 and macOS.

Test case:

- build some project on esp32 launch target, then switch target on esp32s2, click "Yes" in IDF target changed dialog. After that, click the Build button -> project is successfully built.

### DIFF
--- a/bundles/com.espressif.idf.core/src/com/espressif/idf/core/util/SDKConfigJsonReader.java
+++ b/bundles/com.espressif.idf.core/src/com/espressif/idf/core/util/SDKConfigJsonReader.java
@@ -63,16 +63,20 @@ public class SDKConfigJsonReader
 		}
 
 		JSONParser parser = new JSONParser();
+		FileReader reader = new FileReader(sdkconfigJsonPath);
 		try
 		{
-			return (JSONObject) parser.parse(new FileReader(sdkconfigJsonPath));
-
+			return (JSONObject) parser.parse(reader);
 		}
 		catch (
 				IOException
 				| ParseException e)
 		{
 			throw new Exception(e);
+		}
+		finally 
+		{
+			reader.close();
 		}
 	}
 
--- a/bundles/com.espressif.idf.launch.serial.ui/src/com/espressif/idf/launch/serial/ui/internal/NewSerialFlashTargetWizard.java
+++ b/bundles/com.espressif.idf.launch.serial.ui/src/com/espressif/idf/launch/serial/ui/internal/NewSerialFlashTargetWizard.java
@@ -163,7 +163,7 @@ public class NewSerialFlashTargetWizard extends LaunchTargetWizard {
 				deleteDirectory(file);
 			}
 		}
-		return directoryToBeDeleted.delete();
+		return directoryToBeDeleted.getName().equals("build") ? true : directoryToBeDeleted.delete();
 	}
 
 	private void cleanSdkConfig(IResource project) {


bundles/com.espressif.idf.launch.serial.ui/src/com/espressif/idf/launch/serial/ui/internal/NewSerialFlashTargetWizard.java
```diff
@@ -163,7 +163,7 @@ private boolean deleteDirectory(File directoryToBeDeleted) {
 				deleteDirectory(file);
 			}
 		}
-		return directoryToBeDeleted.delete();
+		return directoryToBeDeleted.getName().equals("build") ? true : directoryToBeDeleted.delete();
```
kolipakakondal, line 166: Hi @sigmaaa 

I am not quite sure how this is helping here. Are we get into a case where we pass the non "build" directory to this method?

Could you replace this File delete recursive implementation with the Files.walkFileTree(..) and see if that helps here

```
Files.walkFileTree(directoryToBeDeleted, new SimpleFileVisitor<Path>() {
		   @Override
		   public FileVisitResult visitFile(Path file, BasicFileAttributes attributes) throws IOException {
		       Files.delete(file); 
		       return FileVisitResult.CONTINUE;
		   }

		   @Override
		   public FileVisitResult postVisitDirectory(Path dir, IOException exc) throws IOException {
		       Files.delete(dir); 
		       return FileVisitResult.CONTINUE;
		   }
		});
```

bundles/com.espressif.idf.core/src/com/espressif/idf/core/util/SDKConfigJsonReader.java
```diff
@@ -63,17 +63,21 @@ protected JSONObject read() throws Exception
...
 		}
 		catch (
 				IOException
 				| ParseException e)
 		{
 			throw new Exception(e);
 		}
+		finally 
+		{
+			reader.close();
```
kolipakakondal, line 79: Good catch

bundles/com.espressif.idf.launch.serial.ui/src/com/espressif/idf/launch/serial/ui/internal/NewSerialFlashTargetWizard.java
sigmaaa: Hi @kolipakakondal,

the problem in the delete function is that we are trying to delete the "build" directory and for some reason when we are trying to delete even an empty "build" directory after switching targets, there is a problem caused by locking this directory by Java. Then it's impossible to delete this directory or build the project once again (Only restart eclipse helps). Please check the following gif:

IEP-317-demonstration

To avoid this problem, I added a condition statement, that's checks if we are deleting the "build" directory and returns "true" just to skip deleting. Since we are deleting the "build" directory at the end of the delete process, it will leave an empty build directory at the end of the delete function. This is how it works with this statement: 

IEP-317-demonstration2

As you can see, there is an empty "build" directory after switching targets with no lock on it, so we can build the project with no error.

Once again, this issue happens only on Windows 10.

PR #211, opened 2021-03-02
IEP-42 UE improvments in product log, added null checks

sigmaaa:
User experience improvement in product log: 
Now if some variables in the environment are missing, the user will not get a message dialog to install missing tools, but instead will see the product log, where he can see what variables are missing.

IEP-42_2

Also, was added condition statements to avoid NPE.


kolipakakondal (approved):
LGTM

PR #214, opened 2021-03-04
IEP-378 deleting build folder after rename/copy/import

sigmaaa:
Cleanup the build folder after copy/rename/import as it contains old project artifacts like cmakecache.txt which causes build failure.

Test cases: 
create a new project > build > rename > build renamed project - passed
create a new project > rename > build renamed project - passed
copy some built project > delete old project > build copied project - passed
copy non-built project > build copied project - passed
copy some built project > built copied project - passed (make sure that new project is built)
create new project > build it > delete from workspace > import to a workspace with new name > build it again - passed
create new project > delete from workspace > import to a workspace with new name >build it - passed

tested on Windows 10 and macOS


kolipakakondal (approved):
LGTM

PR #246, opened 2021-04-14
IEP-393: show a list of projects if there is no project selection found

sigmaaa:
The project name field was added to the ESP-IDF Serial monitor
IEP-393


kolipakakondal (approved):
LGTM

PR #253, opened 2021-04-20
IEP-411 add a default value for project 

sigmaaa:
Added a default value for the project field in the create-component wizard.


bundles/com.espressif.idf.ui/src/com/espressif/idf/ui/wizard/NewComponentWizardPage.java
```diff
@@ -112,11 +112,20 @@ public void modifyText(ModifyEvent e)
...
 			projectCombo.add(project.getName());
+			if (optProject.isPresent() && optProject.get().equals(project))
+			{
+				projectCombo.select(index);
+			}
+			else
+			{
+				index += 1;
+			}
+
```
kolipakakondal, line 128: Your old approach should still work without this even if it's ready_only. Do you see any problem with that?

`optProject.ifPresent(project -> projectCombo.setText(project.getName()));`


> /**
>  * Sets the contents of the receiver's text field to the
>  * given string.
>  * <p>
>  * This call is ignored when the receiver is read only and
>  * the given string is not in the receiver's list.
>  * </p>
>  * <p>

kolipakakondal (approved):
LGTM

PR #257, opened 2021-04-27
IEP-409: Handling port closing and opening on serial flash

kolipakakondal:
https://jira.espressif.com:8443/browse/IEP-409

**Test Case:**

1. Build and flash an application
2. Open ESP-IDF serial terminal
3. The serial terminal should connect to the port and show serial output
4. Flash the application again - This should disconnect the port and stop printing the serial output, and once the flashing is over it should connect back to the serial port and start printing the serial output (This was not the case earlier, where we were unable to flash it again until we close the serial monitor window)


sigmaaa (approved):
LGTM. Checked it on windows 10

kolipakakondal:
Thanks @sigmaaa for reviewing.

PR #277, opened 2021-05-26
Fix: load esp_idf.json only once for workspace

kolipakakondal:
This is to avoid issues with the user settings are getting overridden by the esp_idf.json settings every time eclipse launches.


sigmaaa (approved):
LGTM

PR #278, opened 2021-05-27
IEP-389: Target configuration select wrong com port

sigmaaa:
Fixed selecting the wrong port after disconnecting the device.
test case:
connect the device and select the right port to it in the esp target configuration > quit the target configuration wizard > unplug the device and try to flash the project and you will see error msg: 
Screenshot 2021-05-27 at 11 55 17
if we will go to the esp target wizard again, we will see an empty device:
Screenshot 2021-05-27 at 11 56 49

if we connect our device again, we will be able to flash the project again without selecting a port.


kolipakakondal (approved):
LGTM

PR #288, opened 2021-06-07
IEP-457: Missing property finder script

alirana01:
The script is to be run on powershell in the root directory for the idf-eclipse-plugin, it will look for properties files in all directories against the given language locale and generate the output in missing properties folder keeping the same structure


kolipakakondal (changes requested):
Thanks for the @alirana01, could you move this file here https://github.com/espressif/idf-eclipse-plugin/tree/master/common/

Changes LGTM

kolipakakondal (approved):

PR #289, opened 2021-06-07
IEP-461: provide the right path to the board/interface config file and update UI

sigmaaa:
In the latest [openocd](https://github.com/espressif/openocd-esp32/releases/tag/v0.10.0-esp32-20210401)  "esp-config.json" file was added. We are using it to provide to the user the right interface/board in the debugger configurations tab. 
The right path to the esp-config.json was provided and combo elements for voltage, target
 and board now is bigger, so the user is able to see the option he is selecting
 
 Screenshot 2021-06-07 at 15 23 22


kolipakakondal (approved):
LGTM

PR #291, opened 2021-06-10
DOC-1556: Converted from native2ascii

alirana01:
The following files were received from document team and were converted using our own utility the provided ascii conversions seemed off so committing with our own conversions


kolipakakondal (approved):
LGTM

PR #296, opened 2021-06-15
IEP-468: Сustomize the messages according to what is missing

sigmaaa:
different message when tools are not installed from Eclipse, example:

Screenshot 2021-06-15 at 22 02 16

Screenshot 2021-06-15 at 22 07 29

Screenshot 2021-06-15 at 22 09 12


kolipakakondal (approved):
LGTM

PR #304, opened 2021-06-24
IEP-467: Terminal rows and cols now added to config page

alirana01:
The user will be asked for the number of rows or cols for the terminal when connecting. Also updated the call for the pty4j process to pass the given cols and rows

image

### DIFF
--- a/bundles/com.espressif.idf.serial.monitor/META-INF/MANIFEST.MF
+++ b/bundles/com.espressif.idf.serial.monitor/META-INF/MANIFEST.MF
@@ -9,13 +9,15 @@ Require-Bundle: org.eclipse.core.runtime,
  com.espressif.idf.core,
  com.espressif.idf.ui,
  org.eclipse.ui.console;bundle-version="3.9.0",
- org.eclipse.ui
+ org.eclipse.ui,
+ ilg.gnumcueclipse.core
 Bundle-RequiredExecutionEnvironment: JavaSE-11
 Automatic-Module-Name: com.espressif.idf.serial.monitor
 Bundle-ActivationPolicy: lazy
 Export-Package: com.espressif.idf.serial.monitor,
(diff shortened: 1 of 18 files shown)


sigmaaa (approved):
LGTM

kolipakakondal (changes requested):
Thanks @alirana01 for the PR.  Here is what I am thinking on this - 
1.  Can we avoid mixing technical parameters(rows, columns) with the functional parameters in the UI - Instead of showing this in the serial dialog how about moving this to Espressif preferences general page
2. Configuring rows and columns is a one-time activity so it would best fit in the preferences page
3. Can we use user-readable names for these instead of rows and columns - Line width(maximum characters) and Number of lines
4. Can we figure out the default settings for these based on the system screen size
5.  Can you check how the eclipse CDT console is identifying default parameters for these and use the same it the same way

kolipakakondal:
We can develop something like this

Screenshot 2021-06-28 at 9 09 56 AM

alirana01:
> Thanks @alirana01 for the PR. Here is what I am thinking on this -
> 
> 1. Can we avoid mixing technical parameters(rows, columns) with the functional parameters in the UI - Instead of showing this in the serial dialog how about moving this to Espressif preferences general page
> 2. Configuring rows and columns is a one-time activity so it would best fit in the preferences page
> 3. Can we use user-readable names for these instead of rows and columns - Line width(maximum characters) and Number of lines
> 4. Can we figure out the default settings for these based on the system screen size
> 5. Can you check how the eclipse CDT console is identifying default parameters for these and use them in the same way

@kolipakakondal 
image
The changes are done as requested. The eclipse cdt console is also using the default value from inside the code as their own source code clearly shows that they are setting default values from the code into the preference store.
Please see if there are any other changes required

Thanks

kolipakakondal (approved):
LGTM

PR #316, opened 2021-07-06
IEP-487: set up INSTALL_TOOL_FLAG during early startup

sigmaaa:
We need to set INSTALL_TOOLS_FLAG when configuring paths for the first time with the offline installer


kolipakakondal (approved):
LGTM

Issue #309, opened 2021-06-26
Eclipse Plugin: Indexer Fails on Change of Run/Debug Configuration. Requires Eclipse Restart ESP32-C3 (IDFGH-5458) (IEP-479)


BrianAtDocumentedDesigns:
This has also been posted here https://www.eclipse.org/forums/index.php/m/1842646/#msg_1842646 but they suggest I post here as well.

## Environment
Windows 10 Eclipse 2021-03. ESP-IDF current master branch (required because released ESP-IDF is not compatible with ESP32-C3)

- Development Kit: ESP32-C3 mini devkit. 

- Kit version : [v1|v2|v3|v4] I do not know
- 
- Module or chip used:  ESP32-C3 v3
- 
- IDF version :   v4.4-dev-1594-g1d7068e4b (have to use development branch to work with ESP32-C3)

- Build System:   Eclipse

- Compiler version : riscv32-esp-elf-gcc-8.4.0.exe (crosstool-NG 1.24.0.123_64eb9ff) 8.4.0

- Operating System:    Windows 10

- (Windows only) environment type: [MSYS2 mingw32|ESP Command Prompt|Plain Command Prompt|PowerShell].
- ?
- 
- Using an IDE?:   Yes. Eclipse 2021-03 (2021-06 is not compatible with the ESP-32 plugin)

 Power Supply:     USB

## Problem Description

I am using Eclipse CDT 03-2021 with the ESP-IDF Plugin. The plugin does not seem compatible with 06-2021 but the described behaviour is the same. Switching from run to debug configuration or vice-verse results in unresolved includes even though it will build with no errors, suggesting the indexer is not in synch with the source. Reindexing, "clean", etc., does not help. Usually a change in configuration immediately results in index issues but if it does not a re-index or build all always does. This seems to be a common problem.

### Expected Behavior

Indexer should always work. Restart of Eclipse should not be required.

### Actual Behavior
Switching run/debug configuration borks indexer as described

### Steps to reproduce

Start a project (i.e. blink) and create a run and debug configuration for ESP32-C3 (note the default debug configuration does not set up properly for ESP32-C3:

===================
Config options 
-s ${openocd_path}/share/openocd/scripts  -f board/esp32c3-builtin.cfg
Actual Executable
 (path to) \riscv32-esp-elf-gdb.exe 

GDB Client Setup/Other Commands add
         set arch riscv:rv32
==================

[...]

PR #319, opened 2021-07-13
IEP-479: Fixed indexer fails issue on change configuration

sigmaaa:
This PR relates to issues described here: https://github.com/espressif/idf-eclipse-plugin/issues/309

this PR fixes issue with failing indexer on change configuration. 
The second problem was that building in Debug Config fails. It happened because we were losing target after the eclipse restart. This problem was also fixed, now we are not losing target and after eclipse restart, the target bar is not missing for debug config. 

An additional fix for the NPE, when creating a new launch configuration was added. 

Tested on macOS and Windows 10. Test cases:
switching configs, creating a new launch configuration, modify the configuration, delete a configuration, switching between different projects, project close and reopen and eclipse restarting, switching workspaces. 
Expected result: Indexer not failing, building in debug/launch configuration works


kolipakakondal (approved):
LGTM

PR #324, opened 2021-07-22
IEP-490: Provided right gdb executable to gdb client

sigmaaa:
The default option for gdb executable in the GDB setup client is now based on IDF_TARGET


kolipakakondal (approved):
LGTM

PR #333, opened 2021-08-03
IEP-498 Indexer fails on changing the launch target

sigmaaa:
Functionality with the update method was moved to the LaunchBarListener, that's triggering immediately after we are changing the target to the target with a different IDF_TARGET value. Functionality is a little bit refactored, so it's working with debug mode also. 

Test cases:

1.  Create and build the project > change target > "IDF TARGET CHANGED" message appears > click Yes > project is cleaned and the next build is working.
2. Create and build the project > change target > "IDF TARGET CHANGED" message appears > click No > Indexer fails due to the incorrect target 
3. Select some project > create a new launch target > "IDF TARGET CHANGED" message appears if the new launch target is different from one in the build folder configuration
4. Repeat test cases 1,2 and 3 on the debug configuration.


sigmaaa:
added commit related to an indexer issue. Indexer now not failing.
Test cases:
1. Build a project with different targets on the launch configuration
2. Build a project with different targets on the debug configuration
3. Tried on different project templates
4. Change mode from run/debug and build
5. Close/Open > Build
6. Rename > Build
7. Restart eclipse > Build

kolipakakondal (approved):
LGTM

Issue #282, opened 2021-06-02
New Eclipse plugin shows randomly unresolved types, vars members etc. (IEP-444)


samsam4:
**Describe the bug**
New Eclipse plugin shows randomly unresolved types, vars members etc.in Problems View while build output in Console shows always: Build complete (0 errors, 0 warnings)

And the Problems alternate with different amounts between 1 and 16 on every next press of the Build button.

In the original code that I downloaded from [https://github.com/nopnop2002/esp-idf-ili9340](https://github.com/nopnop2002/esp-idf-ili9340)
while trying to solve the problem, I added   #include "dirent.h"   but the output is absolutely same as output without this extra line . The compiler/indexer sure sees the dirent.h, because I tried with slightly modifying the name and get error mark in front of it right away. 
Still cant find why cant see the DIR type defined there (always) and Fields under dirent* pe on every even try to Build (once build no problem with those fields, next build 15 more problems with those fields, next build again no problem with those fields (only DIR type problem - always)?  

The same code compiles, build and flash from CMD environment without any reported problems, this  only happens in Eclipse!

**To Reproduce**
1. Downloads project from here:
[https://github.com/nopnop2002/esp-idf-ili9340/archive/refs/heads/master.zip](https://github.com/nopnop2002/esp-idf-ili9340/archive/refs/heads/master.zip)
2. Import the project
3. Build
(4. )Set sdkconfig for TFT Configuration as per last screenshot
(5. )Build again
6. Observe Console and Problems windows
7. Build again 
8. Observe Console and Problems windows

**Expected behavior**
No errors reported in Problems window or at least same number errors reported on every next Build without modifying anything in code and/or settings.  

**Screenshots**
![Prpblems1](https://user-images.githubusercontent.com/30199901/120408599-5c9a0e00-c31d-11eb-8df0-8261830729f6.png)

![Problems](https://user-images.githubusercontent.com/30199901/120408614-64f24900-c31d-11eb-8c13-5e228b8d8d6c.png)

![Settings](https://user-images.githubusercontent.com/30199901/120408621-67ed3980-c31d-11eb-8159-540f5ac3d499.png)

**Environment Information:** 

Operating System: windows 10

Java Runtime Version: 15.0.2+7-27
Eclipse Version: 4.19.0.v20210303-1800
Eclipse CDT Version: 10.2.0.202012191711
IDF Eclipse Plugin Version: 2.1.0.202104140338
(installed with 2.9beta online installer)

[...]

PR #356, opened 2021-09-06
GH #282: New Eclipse plugin shows randomly unresolved types, vars members etc.

sigmaaa:
Fixed problem, when we are passing dummy headers for the indexer instead of ESP-IDF component's headers.
Test cases:
1. Include some dummy header, that you can find in the xtensa folder, for example "#include "sys/dirent.h" -> build project -> instead of dummy header, you will see a header from esp-idf components folder.
2. Include some header, that is not in esp-idf components folder -> build project -> you will see a header from xtensta folder

### DIFF
--- a/bundles/com.espressif.idf.core/src/com/espressif/idf/core/build/IDFBuildConfiguration.java
+++ b/bundles/com.espressif.idf.core/src/com/espressif/idf/core/build/IDFBuildConfiguration.java
@@ -16,11 +16,11 @@ import java.nio.file.FileVisitResult;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
+import java.nio.file.SimpleFileVisitor;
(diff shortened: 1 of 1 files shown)


alirana01 (approved):
~~LGTM I have also verified changes on windows with my branch and they seem to resolve some issues and for the time being this should be merged with master but I may have to change the method but the issue I am working on may still not be directly linked to this area investigation is still ongoing~~

alirana01:
@sigmaaa @kolipakakondal please hold merge on this I am verifying another small thing

alirana01 (commented):
I think I am missing the point of these changes and doing these changes further causes problem with Indexer we need to collaborate and discuss the issues we both are facing and reach a common ground

kolipakakondal (changes requested):
I would suggest to identify the root problem and fix it so that it can avoid other issues on the indexer

PR #360, opened 2021-09-09
IEP-524: IDF launch target changed message is not showing

sigmaaa:
We need to call `manager.addLaunchTarget(typeId, id);` every time, when we are editing esp IDF target. This allows us to trigger the launch bar listener and show the "IDF launch target changed" message when it's needed. It's also allows us to not duplicate code, that is using in launch bar listener.

By calling `manager.addLaunchTarget(typeId, id);` we are not adding target if it already exists in related hashmap, so it's safe to call it every time when we are editing esp idf target.


alirana01 (approved):
LGTM

Issue #325, opened 2021-07-23
Debug Configuration for ESP32-C3 wrong gdb (IEP-492)


BrianAtDocumentedDesigns:
**Describe the bug**
After installing the eclipse plugin, creating a debug profile points to the wrong gdb executable.
 Actual Executable is 
<user>\.espressif\tools\xtensa-esp32-elf\esp-2021r1-8.4.0\xtensa-esp32-elf\bin\xtensa-esp32-elf-gcc.exe
it should be 
<user>\.espressif\tools\riscv32-esp-elf\esp-2021r1-8.4.0\riscv32-esp-elf\bin\riscv32-esp-elf-gdb.exe

**To Reproduce**
1. Go to Run -> Debug Configurations Select ESP-IDF GDB OpenOCD Debugging
2. Click on New Launch Configuration
3. Scroll down to Actual Executable
4. Note wrong gdb executable

**Expected behavior**
The field should point to riscv32-esp-elf-gdb.exe
The wrong gdb result in profuse errors

**Screenshots**

**Environment Information:** 

*** Features:

**IDF Eclipse Plugin Version:** 
Eclipse > About Eclipse > Installation Details > Installed Software > Search for "espressif"
  Espressif IDF Plugins for Eclipse	2.1.2.202107141350	com.espressif.idf.feature.feature.group	ESPRESSIF SYSTEMS (SHANGHAI) CO., LTD

For v1.1.0 and above: Eclipse > About Eclipse > Click on "espresif" icon

**Eclipse Error log:** 
Window > Show View > Error Log

Please attach the error log as described here https://github.com/espressif/idf-eclipse-plugin#error-log

sigmaaa:
Hi @BrianAtDocumentedDesigns, 

fix for this is available in the latest master and in this [build](https://github.com/espressif/idf-eclipse-plugin/suites/3301595730/artifacts/77183061):
To install, follow the instructions [here](https://github.com/espressif/idf-eclipse-plugin#installPluginsUsingLocalFile) .

Best regards,
Denys

BrianAtDocumentedDesigns:
Perhaps I am doing it wrong but I uninstalled the Espressif plugin downloaded the link you provided and install that and I get the same result

Actual executable 
C:\Users\bjpic\.espressif\tools\xtensa-esp32-elf\esp-2021r1-8.4.0\xtensa-esp32-elf\bin\xtensa-esp32-elf-gdb.exe

The file which downloads is 
[...]

PR #368, opened 2021-10-04
IEP-535 Default and saved values for JTAG combo buttons

sigmaaa:
Hi @kolipakakondal, @alirana01 

PR contains:
- the default value for the board combo button (first from the list)
- saving values for jtag combo buttons. 
- changing the launch target, based on the IDF-target selected in the combo button and previously selected serial port.
-  handled the case, where the user doesn't have a launch target with according IDF-target. In this case. We are navigating him to IDF-Target creation wizard:

Screenshot 2021-10-04 at 13 55 22

- for the debug configuration, we are also changing the gdb executable based on what target is selected. 
- Fixed one bug related to gdb executable, it's relates to already [closed issue](https://github.com/espressif/idf-eclipse-plugin/issues/325), but from the comments I see, that users still facing it


sigmaaa:
Test cases:

both on the launch and debug configurations
- select idf target -> board is selected automatically
- delete targets with esp32c3 idf target-> edit configuration -> select esp32c3 target -> IDF Launch Target not found appears:  
Screenshot 2021-10-04 at 13 55 22

- after clicking "Yes" you should be able to create launch target
- change target in combo button -> after changing some target, the target in launch Bar is also changing appropriately
- after editing configuration click OK and open configuration again -> we are getting saved (not default) values for each configuration.

Test cases for the debug configuration:
- change the value for target combo button -> Actual executable for gdb is correct and changing each time after we are changing target value

sigmaaa:
I'm preparing a commit with some refactoring. It will be pushed soon.

PR #377, opened 2021-10-18
Fix: Pass Build env PATH to CMake build

kolipakakondal:
This will make sure PATH from the `Preferences > C/C++ > Build > Environment > PATH` is honored while compiling a project.

**What was happening earlier?**
Eclipse uses the CMake command to build a project and this is run through Java process runner. By default, the PATH environment for the Java process is system PATH. If there are multiple python environments found in the system there was an issue during the compilation stage especially while running on Apple M1 systems due to different python architectures

**What will happen with PR changes?**
This will make sure Java CMake build process is run through `C/C++ > Build > Environment` `PATH` variable. PATH in preference build environment always configured from the the idf.py export command and it has correct `python_env` based on the idf version and system architecture


sigmaaa (approved):
LGTM

alirana01 (approved):
LGTM 👍

kolipakakondal:
As this is fixed with the other PR, closing this issue.

Issue #346, opened 2021-08-24
Eclipse on Windows shows error markers for cpp file but not for c file when I include freertos/FreeRTOS.h but build succeeds anyway (IDFGH-5757) (IEP-513)


davidlehrian:
## Environment
- Development Kit:      ESP32-DevKitC
- Kit version DevKitC: v4
- Module or chip used: ESP32-WROOM-32D
- IDF version: v4.4-dev-2594-ga20df743f
- Build System: CMake 3.20.3
- Compiler version: esp-2021r1-8.4.0
- Operating System:  Windows 10
- Using an IDE?: Yes Eclipse 
  Operating System: windows 10
  Java Runtime Version: 15.0.1+9
  Eclipse Version: 4.20.0.v20210611-1600
  Eclipse CDT Version: 10.3.0.202104042017
  IDF Eclipse Plugin Version: 2.2.0.202108110455
  ESP-IDF v4.4-dev-2594-ga20df743f
  Python set for IDF_PYTHON_ENV: Python 3.8.10
  Python declared in PATH variable: Python 3.9.5
- Power Supply:   USB
- 
## Problem Description

Eclipse IDE is showing an error marker in a cpp file but the same code in a c file does not produce the error.  Both files compile without error.  

### Expected Behavior

I should not be seeing an error marker in the cpp file.  

### Actual Behavior

I am seeing an error marker in the cpp file.  

### Steps to reproduce

1. Create a new project from the sample_project template.
2. Create the following test_function.c and test_function.h files.
```
/* this is a test file */

#include <esp_system.h>

void test_function_a();
```
```
#include <freertos/FreeRTOS.h>
#include <driver/gpio.h>
#include "test_function.h"


void test_function_a(){
	gpio_pad_select_gpio(26);
	gpio_set_direction(26, GPIO_MODE_OUTPUT);
	gpio_set_level(26,0);
}
```
3. Compile the project and notice a clean compile and there are no error markers in Eclipse IDE.
4. Create the following TestClass.cpp and TestClass.h files

```
/* test class */
#include <esp_system.h>

#ifndef CLASS_TestClass_H_
#define CLASS_TestClass_H_

class TestClass {
private:
	TestClass();
	uint8_t GPIO;
public:
	TestClass(uint8_t pin);
	void testMethod();
};

#endif
```
```
/* test class */
#include <freertos/FreeRTOS.h>
#include <driver/gpio.h>
#include "TestClass.h"

[...]

PR #380, opened 2021-10-20
IEP-513 GH #346: Eclipse on Windows shows error markers for cpp file but not for c file when I include freertos/FreeRTOS.h but build succeeds anyway

sigmaaa:
Hi @kolipakakondal, @alirana01.

these changes are fixing this problem https://github.com/espressif/idf-eclipse-plugin/issues/346.
Also fixing some problems with the indexer on project examples, where the indexer was confused because he couldn't get some ESP-IDF headers when they were included as local headers. 

As an example, you can use the "wave_gen" project. 
Before -> indexer couldn't resolve #include "driver/timer.h" inclusion
Now -> Problem is gone. You can navigate to the definition of it and the project has 0 errors in the text editor either.


alirana01 (approved):
LGTM
Just need to verify it over a few more templates

PR #397, opened 2021-11-08
IEP-562 Fix: Avoid appending to existing PATH env variable in CDT build

kolipakakondal:
We are having `python dependencies are not satisfied` issue with the Bosch customer. This has happened because they are having multiple versions of python versions installed and CMake build is picking up the python version from the system instead of python from the CDT build PATH.

Let's rely only on `idf.py export` command `PATH` output for building idf projects. That is having all the required paths and also don't see a need to append to the existing path. That basically avoids configuring multiple tool chains on the path environment also.


sigmaaa:
LGTM

kolipakakondal:
Thanks @sigmaaa for reviewing

PR #401, opened 2021-11-09
IEP- 564: Revert indexer fixes which are causing instability

kolipakakondal:


kolipakakondal:
Hi @alirana01 @sigmaaa I have to revert these changes as I see it's unstable and causing performance issues. Will go for release without these changes and then will pick up these items to work on. We can plan these issues for the next release.

Issue #388, opened 2021-10-27
Atomics are not recognized (IEP-554)


dizcza:
**Describe the bug**
`stdatomic.h` functions and macros are shown in red in Eclipse with the IDF plugin. I'm not sure if this is an issue in the Eclipse itself or specific to the ESP-IDF plugin.

**To Reproduce**
```c
#include <stdatomic.h>
atomic_int a_handle;
```

**Expected behavior**
Eclipse recognizes the `atomic_int` keyword.

**Screenshots**
![Screenshot from 2021-10-27 13-10-24](https://user-images.githubusercontent.com/7688337/139046090-b89fda5d-2a94-477b-b5ba-921af94f6cef.png)

**Environment Information:** 
[eclipse_info.txt](https://github.com/espressif/idf-eclipse-plugin/files/7425227/eclipse_info.txt)

**ESP-IDF Tools Information:**
```
* xtensa-esp32-elf: Toolchain for Xtensa (ESP32) based on GCC
  - esp-2021r2-8.4.0 (recommended, installed)
* xtensa-esp32s2-elf: Toolchain for Xtensa (ESP32-S2) based on GCC
  - esp-2021r2-8.4.0 (recommended, installed)
* xtensa-esp32s3-elf: Toolchain for Xtensa (ESP32-S3) based on GCC
  - esp-2021r2-8.4.0 (recommended, installed)
* xtensa-clang: LLVM for Xtensa (ESP32, ESP32-S2) based on clang (optional)
  - 12.0.1-d9341b81fc (recommended)
* riscv32-esp-elf: Toolchain for 32-bit RISC-V based on GCC
  - esp-2021r2-8.4.0 (recommended, installed)
* esp32ulp-elf: Toolchain for ESP32 ULP coprocessor
  - 2.28.51-esp-20191205 (recommended, installed)
* esp32s2ulp-elf: Toolchain for ESP32-S2 and ESP32-S3 ULP coprocessors
  - 2.28.51-esp-20191205 (recommended, installed)
* cmake: CMake build system (optional)
  - 3.20.3 (recommended, installed)
* openocd-esp32: OpenOCD for ESP32
  - v0.10.0-esp32-20210902 (recommended, installed)
* ninja: Ninja build system (optional)
  - 1.10.2 (recommended, installed)
```

**IDF Eclipse Plugin Version:** 
Version: 2.2.0.202108131352

**Eclipse Error log:** 
[...]

PR #402, opened 2021-11-10
IEP-47: Linking components and indexer fixes

kolipakakondal:
**Current issues:**
- NPE while indexing in macOS - **DONE**
- Components source files (.c) are not linked/added to the esp-idf-components folder in macOS - **DONE**
- Include paths should have an absolute path. It seems to be we are passing relative paths and the indexer is ignoring them. For example `helloworld/ESP_IDF_Components/newlib/platform_include` was reported in the parser log - **DONE**
- Blink example has 5648 unresolved inclusion errors. - **Partially DONE**
  - If we enable `Allow heuristic resolution of Includes` on the project level, this would bring down errors from `5648` to `1457`. However, this seems to be inconsistent and needs deeper investigation.- **No more required**
  - `Allow heuristic resolution of Includes` is getting reset to false after eclipse restart - issue from Eclipse
  - After adding only .c files with our new approach we could bring it down to 7 unresolved header issues. - **DONE**
  -   The remaining unresolved header issues are related to `/esp-idf/examples/...` - **In Progress**
- Remove system includes from local include paths. Adding system includes in the local include file list is not helping anyway. I could see the same number of unresolved inclusion errors  - **DONE**
- Re-visit eclipse linked resource vs symlink resource considering platform dependency - **DONE**
- For every build re-indexing is happening  -**DONE**
- Reduce the number of project refresh calls on resource change. For example`project.refreshLocal` is called for every folder and file creation -  **DONE**
-  There is confusion around merging system includes and local includes in the CMakeInfoConsumer -  **DONE**
- Project is not getting refreshed immediately after build, the user has to wait till the components get added and index them and which is taking a long time
- `.hpp` files are not linked to the project which is causing unresolved headers `#include "nvs.hpp"` -  **DONE**
- `sourceFileLocalIncludesMap` usage? -  **DONE**
- Add a filter list to ignore certain types of files: for example, *.mk, *.txt, *.yml, *.md, etc are also getting created. I don't think we need to link them -  **DONE**
- We could avoid creating a physical folder
- Need to find a way to avoid reporting `ESP_IDF_Components` errors on the project level even if there are certain header unresolved issues to avoid a bad experience for the user
- Indexing after project > C/C++ > Rebuild working differently compared with indexing after general build. Could be because we are adding extra include paths IDFBuild configuration
- On Eclipse restart every project in the workspace is re-indexing -   **DONE**
- On Eclipse restart all linked files headers are unresolved -   **DONE***

**Symbols**
- esp_eth.c: atomic_int ref_count;  - Type 'atomic_int' could not be resolved
- static spinlock_t global_extram_lock = SPINLOCK_INITIALIZER; - Symbol 'SPINLOCK_INITIALIZER' could not be resolved
- -i2c.c Symbol 'SPINLOCK_INITIALIZER' could not be resolved

Indexed 'blink' (879 sources, 1,977 headers) in 48.6 sec: 177,258 declarations; 1,006,471 references; 0 unresolved inclusions; 328 syntax errors; 213 unresolved names (0.018%)


kolipakakondal:
Current issues:
- All headers are not getting resolved on the hello-world project also (macOS)
- For every build re-indexing is happening
- Reduce the number of refresh calls on resource change
-

kolipakakondal:
Hi @sigmaaa  @alirana01 Please review this PR and comment if you see any issues.

alirana01 (approved):
LGTM

igrr:
Have tested this locally and so far the indexer works remarkably well! Great job @kolipakakondal!
The only issue i ran into was https://github.com/espressif/idf-eclipse-plugin/issues/388.

kolipakakondal:
Thanks @alirana01 and @igrr for reviewing. Will check #388 as well.

sigmaaa:
I was looking into code changes and functionality. LGTM!

PR #405, opened 2021-11-12
IEP-574: Fix for update site installation on Eclipse 4.19

kolipakakondal:
Here is the problem:
Screenshot 2021-11-12 at 1 16 27 PM

mentioning a specific version for LaunchBar plugin is causing this issue.


sigmaaa (approved):
LGTM

PR #411, opened 2021-11-30
IEP-584: Logic fixed in finding the path from the system path

alirana01:
### Win OS: Git executable location is not found

The class `ExecutableFinder` was returning the match in find method for only the first path and if found it was okay if not it was null which it was for majority cases. Fixed that by adding null check and continuing the loop till it finds the path.

image


kolipakakondal (approved):
Good catch. How did I miss that!!

PR #422, opened 2021-12-22
IEP-600: ESP-IDF Components feedback

alirana01:
Fixes for NPE, SWT Resource exceptions and installations issues. The search functionality is remaining. The UI is also now changed back to generic colors of the existing UI


kolipakakondal (approved):
LGTM

PR #428, opened 2021-12-30
IEP-542 Serial port was not found error while JTAG flashing

sigmaaa:
The "Serial port not found" error does not occur when the user selects "flash over jtag" in the launch configuration.
But I decided to edit this error message anyway to make it more user-friendly. Instead of showing an error, we now guide the user to edit their launch target, where they can select a serial port.
How to reproduce:

1. Create the launch target without selecting the serial port
2. Flash the project over UART
3. You will see confirm dialog, click OK
Screenshot 2021-12-30 at 11 46 02
4. Now you will see the "Edit launch target" dialog, where you can select the Serial Port
Screenshot 2021-12-30 at 11 47 00


kolipakakondal (approved):
LGTM

PR #434, opened 2022-01-11
IEP-616: Flash over JTAG checkbox is not visible with the latest eclipse and cdt

sigmaaa:
On the latest eclipse with the latest CDT, "flash over Jtag" checkbox in the launch configuration is not visible
Eclipse Version: 4.22.0.v20211124-1800
Eclipse CDT Version: 10.5.0.202109270023

To fix this, I had to move this checkbox to the main composite


kolipakakondal (approved):
Thanks @sigmaaa for the PR. LGTM

Do all these methods exist in the Eclipse 2021-06 and 2021-09 also? Just to make sure it's also working with older eclipse releases

sigmaaa:
Hi @kolipakakondal,
thanks for the review. I also checked these changes on older versions 06 and 09 and it's working.

PR #435, opened 2022-01-12
IEP-602: Support spaces in IDF_PATH and project path

sigmaaa:
PR summary:

- Allowing spaces in IDF_PATH if a selected version is equal to or higher than 5.0
- If existing ESP-IDF is selected, then we are showing confirming message:
spaces detected
- Disabling spaces check for project creation with esp-idf 5.0 and higher.
- Replacing " " with "_" in places, when we are looking for elf, bin files based on the project name.
- Added the following code to the CMakeLists.txt:
`get_filename_component(ProjectId ${CMAKE_CURRENT_LIST_DIR} NAME)
string(REPLACE " " "_" ProjectId ${ProjectId})
project(${ProjectId})`
With this, binaries after the build will be always as project name directory. If the project name has spaces, they will be replaced with "_". So for "Hello world" project, we will have "Hello_world.bin" file in build directory and etc.


kolipakakondal:
Good work @sigmaaa

PR #439, opened 2022-01-17
IEP-547 OpenOCD: Provide an option for enabling OpenOCD verbose output

sigmaaa:
Hi @kolipakakondal, @alirana01. With the `-d3 -l oocd.log` openocd command our debugger is not working properly, target is not being halted. To fix it, I had to enable verbose output without saving it to the log, just by adding debug level 3 - '-d3'. Since eclipse has the option to save the debug log to a file, I suggest using that instead of '-l oocd.log'. With this option, debugging works fine.

Option to enable verbose output is added to the startup tab:
verbose output

To save log into a file:
save log

Also, it's necessary to update the jtag debugging documentation, I will do it soon in the next commit to this PR.


kolipakakondal (approved):
Thanks @sigmaaa for the PR. LGTM

It would be great if you could update the documentation on this in the OpenOCD debugging doc page https://github.com/espressif/idf-eclipse-plugin/blob/master/docs/OpenOCD%20Debugging.md

Also, add a new screenshot

kolipakakondal (approved):
LGTM

PR #441, opened 2022-01-18
IEP-621 ESP-IDF version check breaks IDF download page

sigmaaa:
Based on the https://github.com/espressif/esp-idf/tags. Tags could contain patterns like `-dev`, `-beta`, `-rc1`. So we are replacing `-.*` regex pattern with 0 to check if the version is higher than 4.0.0


alirana01 (approved):
LGTM! Thank you

kolipakakondal (approved):

PR #443, opened 2022-01-19
IEP-620: Changes for the file to be copied to rules directory for linux

alirana01:


bundles/com.espressif.idf.ui/src/com/espressif/idf/ui/update/InstallToolsHandler.java
```diff
@@ -59,13 +68,16 @@ protected IStatus run(IProgressMonitor monitor)
...
+
 				monitor.setTaskName(Messages.InstallToolsHandler_AutoConfigureToolchain);
 				configureToolChain();
 				monitor.worked(1);
 				console.println(Messages.InstallToolsHandler_ConfiguredCMakeMsg);
-				
+
 				console.println(Messages.InstallToolsHandler_ToolsCompleted);
+
+				copyOpenOcdRules();
```
kolipakakondal, line 79: Can we place this code right after autoconfiguring the tools step? 
```
configureToolChain();
monitor.worked(1);
copyOpenOcdRules();
```

bundles/com.espressif.idf.ui/src/com/espressif/idf/ui/update/InstallToolsHandler.java
```diff
@@ -83,6 +95,38 @@ protected IStatus run(IProgressMonitor monitor)
 		installToolsJob.schedule();
 
 	}
+	
+	private void copyOpenOcdRules()
+	{
+		if (Platform.getOS().equals(Platform.OS_LINUX))
+		{
+			if (!IDFUtil.getOpenOCDLocation().equalsIgnoreCase(StringUtil.EMPTY))
```
kolipakakondal, line 103: Can we combine these two if statements to avoid nested if statements?

```
if (Platform.getOS().equals(Platform.OS_LINUX)) && !IDFUtil.getOpenOCDLocation().equalsIgnoreCase(StringUtil.EMPTY)) 
{
 //copy rules
}
```

bundles/com.espressif.idf.ui/src/com/espressif/idf/ui/update/InstallToolsHandler.java
```diff
@@ -83,6 +95,38 @@ protected IStatus run(IProgressMonitor monitor)
...
+				pathToRules.append(IDFUtil.getOpenOCDLocation());
+				pathToRules.append("/../share/openocd/contrib/60-openocd.rules"); //$NON-NLS-1$
+				File rulesFile = new File(pathToRules.toString());
+				if (rulesFile.exists())
+				{
+					Path source = Paths.get(pathToRules.toString());
+					Path target = Paths.get("/etc/udev/rules.d/60-openocd.rules"); //$NON-NLS-1$
+					try
+					{
+						Files.copy(source, target, StandardCopyOption.REPLACE_EXISTING);
```
kolipakakondal, line 117: Could you print the source to the destination copy statement with paths in the IDF Console? If something goes wrong while copying the file due to some permissions, the user could try to do it manually if knew what we are trying to do.

kolipakakondal (changes requested):
LGTM

bundles/com.espressif.idf.ui/src/com/espressif/idf/ui/update/InstallToolsHandler.java
alirana01: Done

bundles/com.espressif.idf.ui/src/com/espressif/idf/ui/update/InstallToolsHandler.java
alirana01: Done

bundles/com.espressif.idf.ui/src/com/espressif/idf/ui/update/InstallToolsHandler.java
alirana01: Done

kolipakakondal (approved):
LGTM

gerekon:
LGTM

gerekon:
One thing I'd take into account. Warn user when the file is about to be overwritten. They can have their own custom version of that file installed already. 

BTW, the same belongs to VsCode. cc @brianignacio5

alirana01:
> One thing I'd take into account. Warn user when the file is about to be overwritten. They can have their own custom version of that file installed already. And let users choose should we need to overwrite the file or skip copying.
> 
> BTW, the same belongs to VsCode. cc @brianignacio5

MessageBox added for the user to intimate them about the file if present at the destination

kolipakakondal (approved):

PR #466, opened 2022-02-01
can not create launch configuration ESP-IDF GDB OpenOCD Debugging (IEP-630)

sigmaaa:
The problem occurs in an older version of openocd, which does not support flashing jtag. Fixed by adding null checking


kolipakakondal (approved):
LGTM

PR #469, opened 2022-02-02
Fix: Removing requirements.txt validation check

kolipakakondal:
Removing `requirements.txt` additional validation check as part of the esp-idf folder validation. Will rely on `/<esp-idf folder>/tools/idf.py` file check to validate whether it's an esp-idf directory or not.

Will keep this for v2.4.1 release


sigmaaa (approved):
LGTM

PR #475, opened 2022-02-07
IEP-636: Two versions of the toolchain for each target after installing the new esp-idf and tools.

sigmaaa:
added: removing previously installed toolchains after installing new tools.

before changes:
build error and toolchains

after changes:
Toolchains after fix


kolipakakondal (approved):
LGTM

PR #476, opened 2022-02-07
Opening a terminal for the first time after Eclipse starts hangs at "Connecting..." and throws an NPE (IEP-637)

sigmaaa:
sometimes socket server starts after the SocketServerHandler.getServerPort(), because the Socket server starts in a different thread. These changes are forcing to start socket server first


kolipakakondal (approved):
LGTM

PR #484, opened 2022-02-15
openOCD can not launch debug session but can flash using JTag (IEP-643)

sigmaaa:


kolipakakondal (approved):
LGTM. I think we need to verify this in all environments with different scenarios.

PR #487, opened 2022-02-16
Version 2.4.1 (Eclipse) and 2.4.0 (Espressif IDE): Build fail after image is build for working project without clear reason. (IEP-647)

sigmaaa:
Made code for checking the remaining size more robust. Works if the binary name and project name do not match and if a binary file is missing

### DIFF
--- a/bundles/com.espressif.idf.core/src/com/espressif/idf/core/util/ParitionSizeHandler.java
+++ b/bundles/com.espressif.idf.core/src/com/espressif/idf/core/util/ParitionSizeHandler.java
@@ -37,10 +37,11 @@ public class ParitionSizeHandler
 	// checking the size consists of the idf_size.py command and checking the remaining size from the partition table
 	public void startCheckingSize() throws IOException, CoreException
 	{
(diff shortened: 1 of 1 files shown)


bundles/com.espressif.idf.core/src/com/espressif/idf/core/util/ParitionSizeHandler.java
```diff
@@ -97,14 +100,22 @@ private IPath getMapFilePath(IProject project)
 		return null;
 	}
 	
-	private void checkRemainingSize()
+	private IPath getBinFilePath(IProject project) {
+		GenericJsonReader jsonReader = new GenericJsonReader(project,
+				IDFConstants.BUILD_FOLDER + File.separator + "project_description.json"); //$NON-NLS-1$
+		String value = jsonReader.getValue("app_bin"); //$NON-NLS-1$
```
kolipakakondal, line 106: My bad, I would have suggested this earlier during the review.

kolipakakondal (approved):
LGTM

PR #493, opened 2022-02-22
Espressif preferences for the serial monitor has no effect on it

sigmaaa:
Changed the preference store, to get SERIAL_MONITOR_NUMBER_OF_CHARS_IN_LINE and SERIAL_MONITOR_NUMBER_OF_LINES variables from the espressif preference page. If these changes are ok, we can remove the initializeDefaultPreferences method, since looks like we don't use it anywhere else


kolipakakondal:
Hi @alirana01 Could you verify these changes and check how this was working earlier

alirana01:
> Hi @alirana01 Could you verify these changes and check how this was working earlier

Hi @kolipakakondal I have made some further changes to the PR.
Initially the preferences were stored with the different package it was not working now as we shifted to a different preference store usage when we updated the preferences page for some reason on my local env the store was still showing defaults from the older code. A clean workspace helped 
Please review once the final changes else it LGTM.

kolipakakondal (approved):
LGTM

PR #498, opened 2022-03-02
IEP-634: Bug fix for Language Change

alirana01:
Original issue: Unable to change a language when the root folder name is not eclipse

Fixed
The issue was we were using constant name eclipse while getting install location which was failing as the RCP is now named espressif-ide
We are now using the system property to fetch the name of the launcher and appending that to the path which resolves the error.

Tested on Windows need testing on MAC and LINUX


kolipakakondal (approved):
LGTM

PR #507, opened 2022-03-09
Display sync added for invalid message thread access

alirana01:
User not able to install due to invalid thread access so added the syn call to the default display

### DIFF
--- a/bundles/com.espressif.idf.ui/src/com/espressif/idf/ui/update/InstallToolsHandler.java
+++ b/bundles/com.espressif.idf.ui/src/com/espressif/idf/ui/update/InstallToolsHandler.java
@@ -23,8 +23,8 @@ import org.eclipse.core.runtime.Status;
 import org.eclipse.core.runtime.jobs.Job;
 import org.eclipse.core.runtime.preferences.InstanceScope;
 import org.eclipse.swt.SWT;
+import org.eclipse.swt.widgets.Display;
 import org.eclipse.swt.widgets.MessageBox;
-import org.eclipse.ui.PlatformUI;
 import org.osgi.service.prefs.BackingStoreException;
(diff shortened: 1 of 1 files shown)


kolipakakondal (approved):
LGTM. Thanks for a quick fix. Could you investigate why this was working earlier in your test so that we could avoid it in the future?

alirana01:
> LGTM. Thanks for a quick fix. Could you investigate why this was working earlier in your test so that we could avoid it in the future?

@kolipakakondal still going to test it on Linux for some reason you dont see these Invalid thread access errors on windows when launching within eclipse debugger. To prevent this we must try to test the changes independently from the maven build so I am waiting for runner to complete the action and I will test on Linux VM

alirana01:
@kolipakakondal the fix is tested and fine can be merged at your discretion

kolipakakondal:
In retrospect, after your last change [here](https://github.com/espressif/idf-eclipse-plugin/pull/443#issuecomment-1018322906) related to MessageBox we haven't done verification in linux environment. My bad, I have approved without giving much attention to these changes. You have not seen this issue in Windows considering [this](https://github.com/espressif/idf-eclipse-plugin/blob/1a7fbd3071267ab69300bd2569b318e49ffa7c48/bundles/com.espressif.idf.ui/src/com/espressif/idf/ui/update/InstallToolsHandler.java#L103) call invokes only in Linux platform.

Let's make sure we do this functional test every time before we push and before we approve the PR. Once the UI test cases are up and running, we could further enhance to run all UI/non-UI test cases on various platforms

alirana01:
@kolipakakondal Yes definitely the tests are atleast running on our runner now so I am trying to fix issues that are happening on the VM mostly related to configuration of env and that more wait times are required for them and they will hopefully be ready

PR #514, opened 2022-03-21
IEP-671: ESP-IDF Install Components not fetching the components sources for installed components.

alirana01:
The initial code was not catering for the download of sources and headers for the components. We just need to specify the IDF_COMPONENT_MANAGER variable in the environment and the CMake build system will take care of the rest.
Sources for this change are the documentation from
https://docs.espressif.com/projects/esp-idf/en/latest/esp32/api-guides/tools/idf-component-manager.html

image

 
Also updated the code to use the Eclipse Jobs API to avoid the UI getting stuck on the install operation


kolipakakondal (approved):
LGTM

sigmaaa (approved):
LGTM

PR #516, opened 2022-03-29
IEP-656: Plugin install doesn't install the websocket-client package needed by the terminal

sigmaaa:


bundles/com.espressif.idf.ui/src/com/espressif/idf/ui/update/InstallToolsHandler.java
```diff
@@ -189,5 +199,52 @@ protected void handleToolsInstallPython()
 		arguments.add(IDFConstants.TOOLS_INSTALL_PYTHON_CMD);
 		runCommand(arguments);
 	}
+	
+	protected void handleWebSocketClientInstall()
+	{
+		
+		IPath pipPath = new org.eclipse.core.runtime.Path(pythonExecutablenPath.replace("python.exe", "pip3.exe")); //$NON-NLS-1$ //$NON-NLS-2$
```
alirana01, line 206: I am a bit confused that why are we using exe (extension) which is windows based shouldn't there be a more universal approach to install this or is this issue only related to windows because from what I am seeing in the user reported issue is that the user is using Linux or Unix based OS on the basis of the directory in the given logs

alirana01 (commented):
LGTM just verify on different operating systems

bundles/com.espressif.idf.ui/src/com/espressif/idf/ui/update/InstallToolsHandler.java
sigmaaa: hi @alirana01,

thanks for the review. You are right, should be used more universal approach here. Will fix this in future commit

alirana01 (approved):
LGTM 
Please verify on different Operating Systems just one more time before merge

kolipakakondal (approved):
Verified on Windows and macOS. LGTM.

PR #517, opened 2022-03-29
Fix: Update pty 0.12.7 jar to support Apple M1 arch

kolipakakondal:
Updating pty lib to 0.12.7 jar to fix serial monitor issue due to the pty lib is not loading in the arm64 based systems

Note: This is fixed with https://github.com/JetBrains/pty4j/commit/4fd4f84d701f3418539c59322f5ae674e4e92cec
which is available from pty4j version 0.11.3 or higher

Verification in  0.12.7 jar:
Screenshot 2022-03-29 at 5 05 33 PM


kolipakakondal:
Merging this as we have verified along with @alirana01

PR #520, opened 2022-04-07
IEP-679: Auto change the launch target after clicking on the launch configuration

sigmaaa:
Added a message, when a selected target doesn't match the target, which is selected for the jtag flashing. It will navigate the user to configuration options, where it's possible to change the selected target and a board for the jtag flashing
IEP-678

Test cases:
1. For the launch configuration, select the "Flash over JTAG" option and choose some board with an esp32 chip and click OK. Then select new target esp32s2. After this, you will be able to see this message. Click "Yes" and you should be navigated to the launch target options, where it's necessary to select a new board.
2. Same as 1 test case, but for debug configuration
3. Unselect "Flash over JTAG" option from the launch configuration. In this case, after selecting a target, you should not see same message, but if project is built, you will see a message with a suggestion to delete the build folder.


alirana01 (changes requested):
This LGTM but since we have now added our ui tests to the build via github actions we must try to add newer cases to cover up our use cases. The IEP-679 can be covered with a UI test via swtbot so any new functionality in future does not break this.

If you need help writing a test case the first point of help is the README.md in https://github.com/espressif/idf-eclipse-plugin/tree/master/tests/com.espressif.idf.ui.test 

Let me know if you need help in understanding and recording the tests, you can also have a look at the existing cases.

PR #529, opened 2022-04-28
esp-idf-template: make compatible with IDF 4.1-5.0

igrr:
Issue reported in https://github.com/espressif/esp-idf/issues/8826, template app can't be compiled with `master` branch of ESP-IDF, since the template app uses some APIs which are removed in IDF v5.0.

This PR makes the following changes:

1. Since IDF v3.3 is not supported anymore, update component CMakeLists.txt to use idf_component_register()
2. Simplify template app_main, making it work for IDF v5.0. It is difficult to make sure that a complex template will build for each IDF version. Users can refer to IDF examples for usage of Wi-Fi features.


kolipakakondal:
LGTM. Thanks @igrr for the PR

PR #530, opened 2022-05-04
IEP-692: Fix to restart eclipse for different platforms

alirana01:
Changed to system property **eclipse.launcher** for restarting eclipse.


kolipakakondal (approved):
LGTM

PR #536, opened 2022-05-27
IEP-703: Saving a target attribute for a config each time when editing.

sigmaaa:
We need to save the launch target attribute for the configuration each time we edit it. Otherwise, it may cause the value ​​from the configuration to get out of sync with a value from the launch target bar.

We are already doing it for launch configuration, so this PR contains changes for debug configurations only.


kolipakakondal (approved):
LGTM

PR #542, opened 2022-06-02
IEP-710: Change the message "Project name can't include space" when creating a project

sigmaaa:
the new message is: "Project name can't include space. Spaces in IDF_PATH and project paths are supported in ESP-IDF 5.0 and above". 

Will be appear if the ESP-IDF version is lower than 5.0


alirana01 (approved):
LGTM
**Just take the latest from the master into the branch before merging!**

alirana01:
@sigmaaa just to add to this we can also add a functional test for this as well.

PR #549, opened 2022-06-08
ESP-IDF Terminal: esptool.py is not accessible in terminal

kolipakakondal:
Previously append was removing slash from the path. 

```
Users/kondal/esp/esp-idf-v4.4.1/components/esptool_py/esptool
```

As you could see there was no slash in the front of the path hence esptool.py was not accessible from the terminal. This got fixed now.

**How to verify:**
1. Go to the ESP-IDF terminal
2. type esptoo.py


alirana01 (commented):
Code LGTM!
Will verify with the build locally as well

PR #586, opened 2022-08-19
IEP-681: Show Tools installation error messages in red color

alirana01:
## Description

The idea is to give user the visual feedback on the console when installing the tools. Currently all the installation was shown in the plain default color but now we are showing the error messages in the red color.

Fixes # ([IEP-681](https://jira.espressif.com:8443/browse/IEP-681))

## Type of change

## How has this been tested?

The scripts in the esp-idf directory install_tools.py can be edited to remove or add some lines that can cause the script to fail. This should then return an exception to the process runner which will lead to the status to be error. From there we can capture the exception message and show that error message in red color on the console.

**Test Configuration**:
* ESP-IDF Version:

## Checklist
- [x] PR Self Reviewed
- [x] Applied Code formatting
- [x] Verified on all platforms - Windows,Linux and macOS


alirana01 (commented):
Self Reviewed

kolipakakondal (approved):
LGTM

PR #661, opened 2022-12-13
IEP-814: Fix error for Illegal Thread State exception causing monitor issues

alirana01:
Fix error for Illegal Thread State exception causing monitor issues

## Description

The thread for handling the socket server messages gave IllegalThreadState exception as while flashing the thread was already running and after resume call for the serial port handler the exception was thrown which caused the processes to become orphan and they were showing themselves detach from the eclipse but were running in the background holding the serial port.
Just needed to add a check to see if the thread was alive or not.

Fixes # ([IEP-814](https://jira.espressif.com:8443/browse/IEP-814))

## Type of change

## How has this been tested?
- Flash a project with the monitor running and connected to that project, after flashing the monitor should resume automatically and user must not need to connect again
- Disconnect and reconnect the monitor after flashing.
- Disconnect monitor and flash project and then reconnect monitor

**Test Configuration**:
* OS (Windows):

## Checklist
- [x] PR Self Reviewed
- [x] Applied Code formatting

### DIFF
--- a/bundles/com.espressif.idf.terminal.connector.serial/src/com/espressif/idf/terminal/connector/serial/connector/SerialPortHandler.java
+++ b/bundles/com.espressif.idf.terminal.connector.serial/src/com/espressif/idf/terminal/connector/serial/connector/SerialPortHandler.java
@@ -156,7 +156,11 @@ public class SerialPortHandler
 		};
 
 		thread.start();
-		serverMessageHandler.start();
+		if (!serverMessageHandler.isAlive())
+		{
+			serverMessageHandler.start();
+		}
+
 		isOpen = true;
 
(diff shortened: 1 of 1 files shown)


alirana01 (commented):
minor non breaking change

AndriiFilippov:
@alirana01 
Tested under: 

OS: Windows 10 
ESP-IDF: master 


LGTM.

kolipakakondal:
Hi @alirana01 Please make sure you also add dependency components/features which are getting impacted by this change so that it would help @AndriiFilippov to cover those test cases as well

AFAIK, 
1. Flash
2. Serial Monitor
3. Core dump debugging
4. GDB Stub debugging

PR #662, opened 2022-12-14
IEP-701: Changes to handle the change in idf in esp_idf.json config file

alirana01:
## Description
Changes to handle the change in idf in esp_idf.json config file

Fixes # ([IEP-701](https://jira.espressif.com:8443/browse/IEP-701))

## Type of change

- Bug fix (can be a breaking change)

## How has this been tested?
To test we need to have this tested with installer or some existing espressif ide installation.
- Need to change the idf path in config file
- It should work fine without any changes
- First initialization should be okay

**Test Configuration**:
* ESP-IDF Version:
* 
## Checklist
- [x] PR Self Reviewed
- [x] Applied Code formatting

## For QA 
(@AndriiFilippov)
Use the esp_idf.json file usually found when installing the ide via installer. Try running the IDE and then changing version and IDF_PATH in the json configuration file that's what the installer is doing when used again just to change the version. You should get a message when launching the IDE next time that the version is changed and if yes is pressed the IDE should change the version to the one in the json configuration. 
**Note: Please wait for the tools to complete installation before doing anything in IDE, currently we dont have any progress indicator for that (WIP)**

### DIFF
--- a/bundles/com.espressif.idf.ui/src/com/espressif/idf/ui/InitializeToolsStartup.java
+++ b/bundles/com.espressif.idf.ui/src/com/espressif/idf/ui/InitializeToolsStartup.java
@@ -21,7 +21,9 @@ import org.eclipse.core.runtime.Platform;
 import org.eclipse.core.runtime.preferences.InstanceScope;
 import org.eclipse.launchbar.core.ILaunchBarManager;
 import org.eclipse.osgi.service.datalocation.Location;
+import org.eclipse.swt.SWT;
 import org.eclipse.swt.widgets.Display;
+import org.eclipse.swt.widgets.MessageBox;
 import org.eclipse.ui.IStartup;
 import org.json.simple.JSONObject;
 import org.json.simple.parser.JSONParser;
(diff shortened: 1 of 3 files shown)


bundles/com.espressif.idf.ui/src/com/espressif/idf/ui/InitializeToolsStartup.java
```diff
@@ -165,6 +196,44 @@ public void run()
 		}
 	}
 
+	private void checkForUpdatedVersion(File idf_json_file)
+	{
+		// read esp-idf.json file
+		JSONParser parser = new JSONParser();
+		try
+		{
+			JSONObject jsonObj = (JSONObject) parser.parse(new FileReader(idf_json_file));
```
sigmaaa, line 205: It's better and safer to use try-with-resources with a reader , which implements AutoClosable. The try-with-resources statement ensures that each resource is closed at the end of the statement.

bundles/com.espressif.idf.ui/src/com/espressif/idf/ui/InitializeToolsStartup.java
```diff
@@ -165,6 +196,44 @@ public void run()
...
+			String idfVersionId = (String) jsonObj.get(IDF_VERSIONS_ID);
+			JSONObject list = (JSONObject) jsonObj.get(IDF_INSTALLED_LIST_KEY);
+			if (list != null)
+			{
```
sigmaaa, line 209: if (list ==null ) {return;}  if we want to reduce nesting here

sigmaaa (commented):
Hi Ali, thanks for the PR. We can reduce the nesting in some places, which, in my opinion, improves the readability of the code, and use try-with-resources with an auto-closing reader.  Otherwise LGTM

kolipakakondal (approved):
LGTM. @AndriiFilippov Could we verify from the functional side on various platforms - windows and macOS?

sigmaaa (approved):

AndriiFilippov:
Hi @alirana01 ! 

Tested under: 
OS - Windows 10
ESP-IDF: v4.4 / v5.0

get an info message when launching the IDE next time that the version is changed 
image

PR #677, opened 2023-01-18
IEP-852: Check for websocket dependency before starting serial monitor

alirana01:
## Description

Verify and install websocket before launching idf monitor.
As we have added webscocket plugin dependency with the core dump debugging it was considered as a regression for customers as it's failed after update. 

Fixes # ([IEP-852](https://jira.espressif.com:8443/browse/IEP-852))

## Type of change

## How has this been tested?
Remove the websocket dependency from the command line using the **_pyhton -m pip uninstall websocket_** using the python from the ESP after installing tools in IDE.
Run and Build any sample project and launch serial monitor. The serial monitor should work now as it checks and installs the dependency or will fail with an exception logged in file stating missing dependencies.

**Test Configuration**:
* ESP-IDF Version: Any
* OS (Windows,Linux and macOS): All

- Serial Monitor
All the events linked to monitor including core dump and gdbstub debugging for the moment and normal working of the terminal should be verified as well.

## Checklist
- [x] PR Self Reviewed
- [x] Applied Code formatting

### DIFF
--- a/bundles/com.espressif.idf.serial.monitor/src/com/espressif/idf/serial/monitor/core/IDFMonitor.java
+++ b/bundles/com.espressif.idf.serial.monitor/src/com/espressif/idf/serial/monitor/core/IDFMonitor.java
@@ -92,7 +92,7 @@ public class IDFMonitor
 	{
 		if(!dependenciesAreInstalled())
 		{
-			throw new Exception("Missing Dependencies"); //$NON-NLS-1$
+			throw new Exception("The WebSocket dependency is missing and cannot be installed automatically"); //$NON-NLS-1$
 		}
 		List<String> arguments = commandArgsWithSocketServer();
 		
--- a/bundles/com.espressif.idf.serial.monitor/src/com/espressif/idf/serial/monitor/core/IDFMonitor.java
+++ b/bundles/com.espressif.idf.serial.monitor/src/com/espressif/idf/serial/monitor/core/IDFMonitor.java
@@ -1,7 +1,6 @@
 package com.espressif.idf.serial.monitor.core;
 
 import java.io.File;
-import java.io.IOException;
 import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
@@ -16,12 +15,12 @@ import org.eclipse.core.runtime.Path;
 import com.espressif.idf.core.IDFConstants;
 import com.espressif.idf.core.IDFCorePlugin;
 import com.espressif.idf.core.IDFEnvironmentVariables;
-import com.espressif.idf.core.ProcessBuilderFactory;
 import com.espressif.idf.core.logging.Logger;
 import com.espressif.idf.core.util.GenericJsonReader;
 import com.espressif.idf.core.util.IDFUtil;
 import com.espressif.idf.core.util.SDKConfigJsonReader;
 import com.espressif.idf.core.util.StringUtil;
+import com.espressif.idf.ui.update.InstallToolsHandler;
 
 /**
  * @author Kondal Kolipaka <kondal.kolipaka@espressif.com>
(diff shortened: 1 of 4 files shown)


alirana01 (commented):
Basic changes to the code no flow disturbed

kolipakakondal (changes requested):
Hi @alirana01 Thanks for the PR and for quickly getting it done.

bundles/com.espressif.idf.serial.monitor/src/com/espressif/idf/serial/monitor/core/IDFMonitor.java
```diff
@@ -127,4 +134,67 @@ public Process start() throws IOException
...
+			status = processRunner.runInBackground(arguments, org.eclipse.core.runtime.Path.ROOT, environment);
+			if (status == null)
+			{
+				Logger.log(IDFCorePlugin.getPlugin(),
+						IDFCorePlugin.errorStatus("Unable to get the process status.", null)); //$NON-NLS-1$
+				return false;
+			}
+
+			Logger.log(status.getMessage());
+			return true;
```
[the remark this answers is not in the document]
alirana01: @kolipakakondal I have changed and modified the code to be reusable now

kolipakakondal (approved):
LGTM

alirana01:
@AndriiFilippov you can test this now when you get time

bundles/com.espressif.idf.serial.monitor/src/com/espressif/idf/serial/monitor/core/IDFMonitor.java
```diff
@@ -86,8 +88,12 @@ private String getMonitorBaudRate()
 		return new SDKConfigJsonReader(project).getValue("ESPTOOLPY_MONITOR_BAUD"); //$NON-NLS-1$
 	}
 
-	public Process start() throws IOException
+	public Process start() throws Exception
 	{
+		if(!dependenciesAreInstalled())
+		{
+			throw new Exception("Missing Dependencies"); //$NON-NLS-1$
```
sigmaaa, line 95: According to best practice, we should treat Exception as an abstract class and therefore not use it directly, because it tells us little about what we expect to go wrong. 
Maybe we can change the text to something more informative, like "The WebSocket dependency is missing and cannot be installed automatically"?

bundles/com.espressif.idf.serial.monitor/src/com/espressif/idf/serial/monitor/handlers/SerialMonitorHandler.java
```diff
@@ -38,7 +36,7 @@ public Process invokeIDFMonitor()
 		{
 			return monitor.start();
 		}
-		catch (IOException e)
+		catch (Exception e)
 		{
 			Logger.log(e);
 		}
```
sigmaaa, line 42: What do you think about adding the popup message in this catch, that will tell the user, what is going wrong exactly? We can use text like this: "The WebSocket dependency is missing and cannot be installed automatically". Otherwise, it can be confusing for the user, why the IDF monitor is not working, and maybe not be obvious to check the log

sigmaaa (commented):
Hi @alirana01, PR is LGTM. I wrote 2 notes that could potentially improve the UX, please take a look.

AndriiFilippov:
@alirana01  hi ! 

Tested under: 
OS - Windows 10 / Linux 
ESP-IDF: v5.0 / master
 
Install ESP-IDF / Tools -> uninstall websocket-client -> create / build / flash -> serial monitor -> everything looks good 👍 
LGTM. 
I have clarifying question on debugging yet - will finish testing on Monday

bundles/com.espressif.idf.serial.monitor/src/com/espressif/idf/serial/monitor/handlers/SerialMonitorHandler.java
kolipakakondal: One suggestion - as this is a core/non-UI class, we could refrain from adding UI dialog here. Instead, we can propagate back the error to the handler who is calling and show the error message in the final UI invoker.

bundles/com.espressif.idf.serial.monitor/src/com/espressif/idf/serial/monitor/core/IDFMonitor.java
alirana01: @sigmaaa I have updated the exception message

bundles/com.espressif.idf.serial.monitor/src/com/espressif/idf/serial/monitor/handlers/SerialMonitorHandler.java
alirana01: Can we create a separate ticket for this I can do it in this but I am afraid it may increase the testing scenarios and will also need to change the code in 2 classes SerialConnector.java and SerialPortHandler.java as one is connecting the monitor and the other one is resuming it in case it is interrupted. So it gets to have multiple scenarios for tests like making sure that both sides are handling exceptions from the open method in the later class invoking the monitor.

PR #691, opened 2023-02-06
IEP-869: Default gdb client path is wrong for the new project

sigmaaa:
## Description

We used the same default gdb client executable for all configurations. This was causing incorrect defaults for new projects. 
Instead of providing a default value from the persistence preference, we are getting the default value based on the selected target, or providing the saved value after the debug configuration is already created

Fixes # ([IEP-869](https://jira.espressif.com:8443/browse/IEP-869))

## Type of change

## How has this been tested?
Test 1:
- create a new project
- select some target 
- create a new debug configuration for this project
- the default target and the default gdb client executable must be correct
Test 2:
- repeat test 1, but with a different target
Test 3:
- try to run a debug

**Test Configuration**:
* ESP-IDF Version:
* OS (Windows, Linux, and macOS):

- Debug configuration

## Checklist
- [ ] PR Self Reviewed
- [ ] Applied Code formatting
- [ ] Added Documentation
- [ ] Added Unit Test
- [ ] Verified on all platforms - Windows, Linux and macOS

### DIFF
--- a/bundles/com.espressif.idf.core/src/com/espressif/idf/core/util/IDFUtil.java
+++ b/bundles/com.espressif.idf.core/src/com/espressif/idf/core/util/IDFUtil.java
@@ -110,7 +110,7 @@ public class IDFUtil
 				+ IDFConstants.IDF_SIZE_SCRIPT;
 		return new File(idf_py_script);
 	}
-	
+
 	/**
 	 * @return tools.json file for tools to install
 	 */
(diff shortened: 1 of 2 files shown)


AndriiFilippov:
@sigmaaa hi ! 


Tested under: 
OS - Windows 10 / Linux 
ESP-IDF: master 

Could not see GDB Executable for esp32c2 target. 
Other targets - LGTM. 

image

PR #714, opened 2023-03-10
IEP-900 EBV feedback - automatically configure available ports for debugging

sigmaaa:
## Description

Checking the availability of ports for debugging. If it's not, we get the next available ports.
Fixes # ([IEP-900](https://jira.espressif.com:8443/browse/IEP-900))

## Type of change

## How has this been tested?

- Test 1:
To test it, I created a simple socket server that occupies the port:
```
public class Server {
    public static void main(String[] args) throws IOException {
        int port = 3333;
        ServerSocket serverSocket = new ServerSocket(port);
        System.out.println("Server is listening on port " + port);

        while (true) {
            Socket socket = serverSocket.accept();
            System.out.println("New client connected");
            socket.close();
            System.out.println("Client disconnected");
        }
    }
}

```

Then, try to debug with the default debug settings. In the openocd log, you will see that for gdb connection port 3334 is used instead of 3333, because port 3333 is occupied by a custom server. 

Repeat this test for other ports.
- Test 2: 
instead of starting a custom server, we can check which ports are already in use on our PC and then set that port in the debug configuration. Since the given port is in use, we should see a different port in the openocd log.

**Test Configuration**:
* ESP-IDF Version:

- Debug

## Checklist
- [ ] PR Self Reviewed
- [ ] Applied Code formatting
- [ ] Added Documentation
- [ ] Added Unit Test

### DIFF
--- a/bundles/com.espressif.idf.debug.gdbjtag.openocd/src/com/espressif/idf/debug/gdbjtag/openocd/dsf/Launch.java
+++ b/bundles/com.espressif.idf.debug.gdbjtag.openocd/src/com/espressif/idf/debug/gdbjtag/openocd/dsf/Launch.java
@@ -39,6 +39,7 @@ import org.eclipse.debug.core.model.IProcess;
 import org.eclipse.debug.core.model.ISourceLocator;
 import org.eclipse.embedcdt.debug.gdbjtag.core.dsf.GnuMcuLaunch;
 
+import com.espressif.idf.core.util.PortChecker;
 import com.espressif.idf.debug.gdbjtag.openocd.Activator;
 import com.espressif.idf.debug.gdbjtag.openocd.Configuration;
 import com.espressif.idf.debug.gdbjtag.openocd.ConfigurationAttributes;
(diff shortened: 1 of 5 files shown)


bundles/com.espressif.idf.debug.gdbjtag.openocd/src/com/espressif/idf/debug/gdbjtag/openocd/dsf/Launch.java
```diff
@@ -92,54 +100,72 @@ public void run() {
...
 
-		if (!config.hasAttribute(IGDBLaunchConfigurationConstants.ATTR_DEBUG_NAME)) {
+		if (!config.hasAttribute(IGDBLaunchConfigurationConstants.ATTR_DEBUG_NAME))
+		{
 			DefaultPreferences fDefaultPreferences = Activator.getInstance().getDefaultPreferences();
 			config.setAttribute(IGDBLaunchConfigurationConstants.ATTR_DEBUG_NAME,
 					fDefaultPreferences.getGdbClientExecutable());
 		}
+
+		PortChecker.getAvailablePort(null);
```
kolipakakondal: Can you help to understand why we are making this call?

kolipakakondal (approved):
LGTM. Care should be taken to limit the loop to check the free port. For example, we can set max attempts to 20, if we are unable to get a free port within that limit then the issue should be something else so it's better to report back to the user and unblock the IDE. Otherwise looks good to me

sigmaaa:
Hi @kolipakakondal. Thanks for the review. Added the max attempts of 20 in the last commit

kolipakakondal (approved):
LGTM

kolipakakondal:
Hi @AndriiFilippov Could you please test this?

PR #721, opened 2023-03-22
IEP-907: Adding descriptive port name from the jSerialComm API

alirana01:
## Description

The only name we can get from the api is descriptive port name which may not correspond to the correct device name but is good enough. I have also added a separate thread to update the serial ports info so when a new device gets connected we simply update the serial ports list and also show a message in the information area for that.

Fixes # ([IEP-907](https://jira.espressif.com:8443/browse/IEP-907))

## Type of change

## How has this been tested?
Connect a board and edit see the available ports. Now connect another board the serial ports list will be updated and selecting them will show their descriptive name in the information area as well

## Checklist
- [x] PR Self Reviewed
- [x] Applied Code formatting


sigmaaa (approved):
LGTM

alirana01:
> 

The description will be different the purpose of this is to show whatever description we can from the system available drivers which may not be correct so it is okay

AndriiFilippov:
@alirana01 ok, what about second part ? Do you think it is useful ? 

image

alirana01:
> @alirana01 ok, what about second part ? Do you think it is useful ?
> 
> image

I am not able to reproduce or understand your second part please add some steps that you are using and also tell me about the boards that you are using

AndriiFilippov:
@alirana01 

LGTM 👍

PR #563, opened 2022-07-04
Fix: Add default user workspace

kolipakakondal:
## Description

Set the default user workspace to the /USER_HOME/workspace

Fixes # ([IEP-xxx](https://jira.espressif.com:8443/browse/IEP-xxx))

## Type of change

- Bugfix

## How has this been tested?

Test Case:
- Download Espressif-IDE with this PR
- Observe the workspace selection dialog during the launch, it should set a default workspace to /USER_HOME/workspace

**Test Configuration**:
* ESP-IDF Version: master
* OS (Windows,Linux and macOS):macOS

## Dependent components impacted by this PR:
- Espressif-IDE startup workspace selection dialog

## Checklist
- [x] PR Self Reviewed
- [ ] Applied Code formatting
- [ ] Added Documentation
- [ ] Added Unit Test
- [ ] Verified on all platforms - Windows,Linux and macOS

PR #736, opened 2023-04-04
IEP-908: Adding the code to support the default workspace directory

alirana01:
## Description

The issue with installer where it sometimes references to the idf toolchain directory for workspace should be resolved now the default workspace is now either in the user home directory or is pointed by the **defaultWorkspace** property in the **esp_idf.json** file

Fixes # ([IEP-908](https://jira.espressif.com:8443/browse/IEP-908))

## Type of change

- Can be a breaking change but code reviews should be able to identify them

## How has this been tested?
**Test 1:**
- Launch the espressif ide
- For the first launch the workspace should now point to (USER_HOME/workspace) directory in the workspace selection dialog
**Test 2:**
- Add the esp_idf.json file if trying without the installer in the espressif ide root directory
- Check to see if the **defaultWorkspace** property is present if not add yourself and point to a path
- Launch the espressif ide the default workspace on the first launch should now point to the directory defined in previous step.
**Test3:**
Also make sure to change the workspace to some other directory and try to relaunch the IDE the workspace selection dialog on second start should always point to the last used directory.

**Test Configuration**:
* ESP-IDF Version: Any
* OS (Windows,Linux and macOS): All

## Checklist
- [x] PR Self Reviewed
- [x] Applied Code formatting

### DIFF
--- a/bundles/com.espressif.idf.core/src/com/espressif/idf/core/util/EspIdfJsonParser.java
+++ b/bundles/com.espressif.idf.core/src/com/espressif/idf/core/util/EspIdfJsonParser.java
@@ -7,6 +7,7 @@ package com.espressif.idf.core.util;
 import java.io.File;
 import java.io.FileReader;
 import java.net.URL;
+import java.nio.charset.StandardCharsets;
 import java.text.MessageFormat;
 
 import org.eclipse.core.runtime.Platform;
@@ -62,7 +63,7 @@ public class EspIdfJsonParser
 		
 		
 		JSONParser parser = new JSONParser();
-		try (FileReader fr = new FileReader(idf_json_file))
+		try (FileReader fr = new FileReader(idf_json_file, StandardCharsets.UTF_8))
 		{
 			JSONObject jsonObj = (JSONObject) parser.parse(fr);
 			gitExecutablePath = (String) jsonObj.get(GIT_PATH);
--- a/bundles/com.espressif.idf.core/src/com/espressif/idf/core/util/EspIdfJsonParser.java
+++ b/bundles/com.espressif.idf.core/src/com/espressif/idf/core/util/EspIdfJsonParser.java
@@ -62,9 +62,9 @@ public class EspIdfJsonParser
 		
 		
 		JSONParser parser = new JSONParser();
-		try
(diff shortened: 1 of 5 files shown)


alirana01 (commented):
Overriding the default IDEApplication class and only the required methods that still required some internal code to be rewritten. I could not found a workaround to set the workspace via configs. Other methods only allow setting the default workspace but that also doesnt work at times. 
I have also refactored and isolated some code that could be reused in multiple classes

bundles/com.espressif.idf.core/src/com/espressif/idf/core/util/EspIdfJsonParser.java
```diff
@@ -0,0 +1,115 @@
...
+			return;
+		}
+		
+		isIdfJsonPresent = true;
+		
+		
+		JSONParser parser = new JSONParser();
+		try
+		{
+			JSONObject jsonObj = (JSONObject) parser.parse(new FileReader(idf_json_file));
```
sigmaaa, line 67: we need to close the FileReader with try-with-resources and provide an explicit character set to the FileReader constructor (e.g. StandardCharsets.UTF_8) to avoid dependency on OS code.

sigmaaa (changes requested):
Hi @alirana01, thanks for the PR, need to close FileReader, otherwise LGTM

bundles/com.espressif.idf.core/src/com/espressif/idf/core/util/EspIdfJsonParser.java
alirana01: @sigmaaa resolved in the latest commit

sigmaaa (approved):
LGTM

kolipakakondal:
Setting the default workspace directory(/user-dir/workspace) in the espressif-ide should be resolved with https://github.com/espressif/idf-eclipse-plugin/pull/563 

There was a bug earlier, and it was fixed with [this](https://github.com/espressif/idf-eclipse-plugin/pull/563/commits/c77b3efbf49b67302a75d30517d3e7bb5258438f#diff-032a5ec5fc6582dfa9a6728b384b1567b5fcbfe34f64b8f25047bdb2fbac2a63L17-L18)

PR #750, opened 2023-04-18
Fix: Set Java 17 as a prerequisite to launch IDE

kolipakakondal:
## Description

Set Java 17 as a prerequisite to launch IDE 

Fixes # ([IEP-776](https://jira.espressif.com:8443/browse/IEP-776))

## Type of change

## How has this been tested?

Test 1:
- Install Java 11
- Launch Espressif-IDE
- It should fail during the launch with an error

Test 2:
- Install Java 17
- Launch Espressif-IDE
- IDE should without an issue

**Test Configuration**:
* ESP-IDF Version: master
* OS (Windows,Linux and macOS): macOS

- Espressif-IDE launch

## Checklist
- [x] PR Self Reviewed
- [x] Applied Code formatting
- [ ] Added Documentation
- [ ] Added Unit Test

### DIFF
--- a/bundles/com.espressif.idf.branding/Espressif-IDE.launch
+++ b/bundles/com.espressif.idf.branding/Espressif-IDE.launch
@@ -34,7 +34,7 @@
     <stringAttribute key="location" value="${workspace_loc}/../runtime-EspressifIDE"/>
     <booleanAttribute key="org.eclipse.jdt.launching.ATTR_SHOW_CODEDETAILS_IN_EXCEPTION_MESSAGES" value="true"/>
     <booleanAttribute key="org.eclipse.jdt.launching.ATTR_USE_START_ON_FIRST_THREAD" value="true"/>
-    <stringAttribute key="org.eclipse.jdt.launching.JRE_CONTAINER" value="org.eclipse.jdt.launching.JRE_CONTAINER/org.eclipse.jdt.internal.debug.ui.launcher.StandardVMType/JavaSE-11"/>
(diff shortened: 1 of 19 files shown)


sigmaaa (approved):
LGTM

AndriiFilippov:
image

kolipakakondal:
>Only one thing : it seems to create workspace folder somewhere in my user folder

This is the expected behavior, this PR also has changes related to setting the default workspace which is /Userhome/workspace

kolipakakondal:
I have reverted changes related to the JRE upgrade for our plugins. Here is why?
https://wiki.eclipse.org/Execution_Environments

> You should seek to use the smallest BREE (Bundle-RequiredExecutionEnvironment) possible. There's several reasons to use a BREE smaller than your prerequisites. One is just sound architecture. Things might change in the future, so you want to code your bundle based on your needs. Maybe that prerequisite will be refactored out next release. Similarly, you do not (normally) have control over your prerequisites BREEs. They might lower theirs, they might raise it ... no reason to hard code some assumption in your bundle, based on your prerequisites.

> Another reason for staying at a lower BREE (or at least not exposing types from a later BREE in your bundle's APIs) is to avoid compilation errors / maintenance overhead in downstream bundles. Example: If one of your API types implements java.lang.Iterable from J2SE-1.5, then this forces all dependent bundles to move to J2SE-1.5 as well -- otherwise, they don't compile any more (although the interface addition may be binary compatible).
```

PR #759, opened 2023-05-05
IEP-946 Test coverage for ExecutableFinder

sigmaaa:
## Description
PR summary: 
* migrating to JUnit5 existing test for IDFVersionReader
* test coverage for ExecutableFinder and the following refactoring
* Made ExecutableFinder abstract so we can create fake objects for further testing and be less coupled to this util class
* Added SystemWrapper interface which makes ExecutableFinder less coupled to System.getenv functionality. It allows us to avoid using PowerMock
* Added test coverage for SystemWrapper as well. 

Fixes # ([IEP-946](https://jira.espressif.com:8443/browse/IEP-946))

## Type of change

## How has this been tested?

- test install tools functionality. We can still find the git, and python paths from the system
- run junit tests

**Test Configuration**:
* ESP-IDF Version:

- Install Tools, 
- Install Tools Wizard

## Checklist
- [x] PR Self Reviewed
- [x] Applied Code formatting
- [ ] Added Documentation
- [x] Added Unit Test
- [x] Verified on all platforms - Windows,Linux and macOS


bundles/com.espressif.idf.core/src/com/espressif/idf/core/ExecutableFinder.java
```diff
@@ -1,106 +1,13 @@
...
-
-		if (Platform.OS_WIN32.equals(Platform.getOS()))
-		{
-			return true;
-		}
-		IFileStore fileStore = EFS.getLocalFileSystem().getStore(path);
-		return fileStore.fetchInfo().getAttribute(EFS.ATTRIBUTE_EXECUTABLE);
-	}
+	IPath find(String executableName, boolean appendExtension);
+	IPath findExecutable(IPath path, boolean appendExtension);
```
sigmaaa, line 12: @kolipakakondal, @alirana01 findExecutable method was originally public, but it looks like we can make it private and don't have it in this interface. Also, I see that we always pass appendExtension as true. Maybe we can rid of it as well?

bundles/com.espressif.idf.core/src/com/espressif/idf/core/SystemExecutableFinder.java
```diff
@@ -0,0 +1,123 @@
...
+
+	public SystemExecutableFinder()
+	{
+		this.systemWrapper = new DefaultSystemWrapper();
+	}
+
+	public SystemExecutableFinder(SystemWrapper systemWrapper)
+	{
+		this.systemWrapper = systemWrapper;
+	}
```
sigmaaa, line 31: I decided to not use a singleton for this utility class because it can be beneficial to not have it singleton for testing and maintenance purposes, also modern JVMs are designed to handle object creation and destruction efficiently, so creating new instances of a stateless utility class should not cause any significant performance issues. 

Also not use the static factory method because the constructors here are clean and I see benefits in the future from being able to have subclasses of this class

sigmaaa (commented):
@kolipakakondal, @alirana01 
FYI: I left some comments to explain some design choices

tests/com.espressif.idf.core.test/src/com/espressif/idf/core/test/IDFVersionReaderTest.java
```diff
@@ -4,16 +4,19 @@
  *******************************************************************************/
 package com.espressif.idf.core.test;
```
sigmaaa, line 5: probably we can move this class to unittest package and leave test package for other kind of tests

bundles/com.espressif.idf.core/src/com/espressif/idf/core/SystemExecutableFinder.java
```diff
@@ -0,0 +1,130 @@
...
+			return false;
+		}
+
+		try
+		{
+			Method m = File.class.getMethod("canExecute"); //$NON-NLS-1$
+			if (m != null)
+			{
+				return (Boolean) m.invoke(file);
+			}
```
sigmaaa, line 117: It looks like we can use file.caExecute() here instead of using reflection, since our plugin requires Java 17 and this method is 100% exists

bundles/com.espressif.idf.core/src/com/espressif/idf/core/SystemExecutableFinder.java
```diff
@@ -0,0 +1,130 @@
...
+		catch (Exception e)
+		{
+		}
+
+		if (isPlatformWindows())
+		{
+			return true;
+		}
+		IFileStore fileStore = EFS.getLocalFileSystem().getStore(path);
+		return fileStore.fetchInfo().getAttribute(EFS.ATTRIBUTE_EXECUTABLE);
```
sigmaaa, line 128: Looks like we are not getting here, because the code above will return file.canExecute() in Java 17

kolipakakondal (approved):
LGTM

tests/com.espressif.idf.core.test/src/com/espressif/idf/core/unittest/ExecutableFinderTest.java
alirana01, line 1: This test class is too big to be self explanatory. I think we can move all the variables and actionable methods inside a private class. This can keep the top of the class very clean and simple and the test cases will make more sense. Considering that its difficult to use GWT approach here we can adapt some areas using Fixture class as we are using in ui functional tests. This will make the code a lot more cleaner and the test cases will be easily added and updated in future. It will also help keep a consistency in design.

alirana01 (changes requested):
I have left a comment that we can probably look into and if needed I can explain on call my idea related to design. The other changes you mentioned I think are valid and can be done just try to take a look at the big test class.

sigmaaa:
Removed an unused method from the ExecutableFinder interface and cleaned up unreachable code

tests/com.espressif.idf.core.test/src/com/espressif/idf/core/unittest/ExecutableFinderTest.java
sigmaaa: Hi, @alirana01, thanks for the review. 
I appreciate the suggestion of using a Fixture class, but I believe it may not be necessary for this unit test class. Allow me to explain my reasoning:

1. The size of the test class is not overly large, and most of the code consists of individual test cases. The setUp method, if used, is relatively concise and optional since it's easy to provide the required state directly in each test method. For example, consider the following test case:
```
	@Test
	void testWindowsReturnsNullWhenSystemReturnsIncorrectExecutable()
	{
		SystemWrapper incorrectSystemWrapper = new SystemWrapper()
		{

			@Override
			public String getPathEnv()
			{
				return foundableTempDir.toString();
			}

			@Override
			public String getEnvExecutables()
			{
				return "/EXE"; //$NON-NLS-1$
			}
		};
		ExecutableFinder executableFinder = new TestableSystemExecutableFinderWindows(incorrectSystemWrapper);

		IPath foundExecutable = executableFinder.find(FOUNDABLE_EXE_FILE_STRING);

		assertNull(foundExecutable);
	}
```
In cases like this, following the GWT structure is straightforward because the ExecutableFinder class is stateless, allowing us to test the result of method execution.
2. Since each test case is isolated, the unit test class does not necessarily need to be self-explanatory as a whole. The descriptive names of the test cases themselves provide sufficient information about their purpose and expected behavior.
3. While a Fixture class can improve readability in functional tests due to their higher level of abstraction and the need for complex setups, it may introduce unnecessary verbosity in unit tests. Unit tests focus on testing small, isolated parts of code with minimal dependencies and ideally no state. Therefore, the introduction of a Fixture class might not provide significant benefits in this context.
4. I agree that maintaining consistency in design is important. However, what if we keep the approach with the Fixture class for functional tests only?

bundles/com.espressif.idf.core/src/com/espressif/idf/core/ExecutableFinder.java
```diff
@@ -1,106 +1,12 @@
...
-import org.eclipse.core.filesystem.EFS;
-import org.eclipse.core.filesystem.IFileStore;
 import org.eclipse.core.runtime.IPath;
-import org.eclipse.core.runtime.Path;
-import org.eclipse.core.runtime.Platform;
-
-import com.espressif.idf.core.util.StringUtil;
 
-public class ExecutableFinder
+public interface ExecutableFinder
```
alirana01, line 9: Would it be better to rename the interface to `IExecutableFinder` for better readability

bundles/com.espressif.idf.core/src/com/espressif/idf/core/SystemWrapper.java
```diff
@@ -0,0 +1,11 @@
+/*******************************************************************************
+ * Copyright 2023 Espressif Systems (Shanghai) PTE LTD. All rights reserved.
+ * Use is subject to license terms.
+ *******************************************************************************/
+package com.espressif.idf.core;
+
+public interface SystemWrapper
```
alirana01, line 7: Same for this one it would be better to have the interface names starting with **I** 
`ISystemWrapper`

alirana01 (commented):
Hi @sigmaaa thank you for explaining I just went over your PR again I am a bit confused about the interfaces you are using and I am not able to get the idea why we need them. About the test cases and this I think we can discuss this on a call it would be better

bundles/com.espressif.idf.core/src/com/espressif/idf/core/ExecutableFinder.java
sigmaaa: "I" prefix is not a standard naming convention in Java for interfaces. Java standard interfaces as a reference:
Map, List, Collection, etc.

bundles/com.espressif.idf.core/src/com/espressif/idf/core/SystemWrapper.java
sigmaaa: "I" prefix is not a standard naming convention in Java for interfaces. Java standard interfaces as a reference:
Map, List, Collection, etc.

sigmaaa:
> Hi @sigmaaa thank you for explaining I just went over your PR again I am a bit confused about the interfaces you are using and I am not able to get the idea why we need them. About the test cases and this I think we can discuss this on a call it would be better

Hi @alirana01. I prefer using interfaces for several reasons, but in this case mostly because of these ones:
1. They leave space for future refactoring/extensibility since we can rely on abstraction and easily replace components in test scenarios.
2. Interfaces make it easier to stub and mock dependencies in your tests. By programming to interfaces, you can create mock implementations or stub implementations of those interfaces, which can be used to simulate specific behaviors and control the responses of the dependencies during testing.

alirana01:
> Hi @alirana01. I prefer using interfaces for several reasons.................

Hi @sigmaaa thanks for the response, I think you are pretty clear on the design and it makes sense to me now in many places. Only thing I would suggest is that my last discussion over stuff when added for future references with @kolipakakondal led to the outcome that we are supposed to keep YAGNI or KISS in mind. If you think it satisfies those things and is viable for not so distant PRs its good to go.  

Happy Testing :)

alirana01:
@sigmaaa about interfaces I can agree that it maynot be a standard convention but it increases the readability by a lot.
Anyway these are some good references I found them helpful
https://stackoverflow.com/a/542097/1722792
https://stackoverflow.com/a/542012/1722792
https://stackoverflow.com/questions/541912/interface-naming-in-java

alirana01 (approved):
Just take a look at the comments mentioned and see if you can incorporate @kolipakakondal should be the final person to finalize design side stuff

kolipakakondal:
Hi @sigmaaa,

Thanks for the PR! I understand that you introduced the `ExecutableFinder` and `SystemWrapper` interfaces for abstraction and flexibility in the design. However, we don't currently have any different implementations for these interfaces.

In general, it's best to only introduce interfaces when you have multiple classes that will implement them, each with a different implementation. This allows us to decouple the different parts of our code and make it more flexible.

If you've designed these interfaces with future extensibility in mind, and you see some use cases coming up where we'll need different implementations, then it's probably fine to leave them as-is. Otherwise, PR looks good to me.

kolipakakondal:
I see builds are getting failed, we can fix and merge this PR. @sigmaaa

sigmaaa:
> Hi @sigmaaa,
> 
> Thanks for the PR! I understand that you introduced the `ExecutableFinder` and `SystemWrapper` interfaces for abstraction and flexibility in the design. However, we don't currently have any different implementations for these interfaces.

Hi @kolipakakondal, thanks for the review. We already using different implementations of SystemWrapper in the tests. I'm planning to use different implementations of ExecutableFinder in future tests for IDFUtil class (but first need to make a dependency injection there). It's possible to mock a class for this purpose instead of a mock interface, but I've heard that some Mockito frameworks don't support mocking classes, so it's better to not rely on it and use an interface for abstraction.

> I see builds are getting failed, we can fix and merge this PR. @sigmaaa

fixed in the last commit

PR #773, opened 2023-05-29
IEP-968 Test coverage for Version class

sigmaaa:
## Description

Test coverage for Version class. In this PR I've tried a different naming convention for test methods with an underscore character. I see an advantage in this type of naming, that we can use `@DisplayNameGeneration(DisplayNameGenerator.ReplaceUnderscores.class)` and it will strip underscores in the test reports, which will make it more readable (If we don't want to specify `@DisplayInfo`) 

Also tried `@DisplayInfo` annotation, which is available in junit5, makes test reports like this:
Screenshot 2023-05-29 at 12 35 54

Fixes # ([IEP-968](https://jira.espressif.com:8443/browse/IEP-968))

## Type of change

## How has this been tested?
run unit tests

**Test Configuration**:
* ESP-IDF Version:

- junit tests

## Checklist
- [ ] PR Self Reviewed
- [x] Applied Code formatting
- [ ] Added Documentation
- [x] Added Unit Test


alirana01 (approved):
LGTM

kolipakakondal (approved):
LGTM

PR #774, opened 2023-05-29
IEP-969 Test coverage for ZipUtility

sigmaaa:
## Description
Test coverage for ZipUtility class
Fixes # ([IEP-969](https://jira.espressif.com:8443/browse/IEP-969))

## Type of change

## How has this been tested?
run junit tests

**Test Configuration**:
* ESP-IDF Version:

junit tests

## Checklist
- [ ] PR Self Reviewed
- [ ] Applied Code formatting
- [ ] Added Documentation
- [ ] Added Unit Test

### DIFF
--- a/tests/com.espressif.idf.core.test/META-INF/MANIFEST.MF
+++ b/tests/com.espressif.idf.core.test/META-INF/MANIFEST.MF
@@ -7,7 +7,8 @@ Bundle-Vendor: Espressif Systems
 Automatic-Module-Name: com.espressif.idf.core.test
 Bundle-RequiredExecutionEnvironment: JavaSE-11
 Require-Bundle: org.junit,
- com.espressif.idf.core;bundle-version="1.0.1"
+ com.espressif.idf.core;bundle-version="1.0.1",
+ junit-jupiter-api
 Bundle-ClassPath: .,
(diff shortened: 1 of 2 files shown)


alirana01 (commented):
Hi @sigmaaa LGTM!
Just one small thing that maybe you can investigate or know already, if zip contains link files as tar.xz or gz we should have some tests for those I remember from the tools installation wizard that some of these zip files that were a link to another file either hard or soft given that file was also present in the archive were getting extracted as zero byte file. Not sure if zip has the same thing or not but worth looking at it. Also maybe we can move that archive decompression in tools installation wizard here. What do you think?

sigmaaa:
> Hi @sigmaaa LGTM! Just one small thing that maybe you can investigate or know already, if zip contains link files as tar.xz or gz we should have some tests for those I remember from the tools installation wizard that some of these zip files that were a link to another file either hard or soft given that file was also present in the archive were getting extracted as zero byte file. Not sure if zip has the same thing or not but worth looking at it. Also maybe we can move that archive decompression in tools installation wizard here. What do you think?

Hi, @alirana01, thanks for the review. I'll take a look into this

kolipakakondal (approved):

PR #775, opened 2023-05-31
IEP-954 In SerialFlashLaunchConfigDelegate we do not rely on the location field

sigmaaa:
## Description

Changed the way how we get python location during launch. Now, instead of value combined with environment python env (from IDFUtil), we are getting it from the location field of the launch configuration.

Fixes # ([IEP-954](https://jira.espressif.com:8443/browse/IEP-954))

## Type of change

## How has this been tested?
Test 1:
- In the launch configuration, change the location field from ${IDF_PYTHON_ENV_PATH} to some other Python path
- flash 
- check logs, you have to see a message like this:
!MESSAGE flash command: /Users/denysalmazov/.espressif/python_env/idf5.1_py3.9_env/bin/python....
make sure that the python path is the same that is selected in the location field
Test 2:
- you can replace ${IDF_PYTHON_ENV_PATH} with ${system_path:python} or similar variable that will find system python path

**Test Configuration**:
* ESP-IDF Version:

- flash

## Checklist
- [x] PR Self Reviewed
- [x] Applied Code formatting
- [ ] Added Documentation
- [ ] Added Unit Test


bundles/com.espressif.idf.launch.serial.core/src/com/espressif/idf/launch/serial/internal/SerialFlashLaunchConfigDelegate.java
```diff
@@ -117,10 +117,12 @@ protected void launchInternal(ILaunchConfiguration configuration, String mode, I
...
 
-		String location = varManager.performStringSubstitution(IDFUtil.getIDFPythonEnvPath());
+		String location = configuration.getAttribute(ICDTLaunchConfigurationConstants.ATTR_LOCATION,
+				SYSTEM_PATH_PYTHON);
+		location = varManager.performStringSubstitution(location);
 		if (StringUtil.isEmpty(location))
 		{
-			location = configuration.getAttribute(ICDTLaunchConfigurationConstants.ATTR_LOCATION, SYSTEM_PATH_PYTHON);
+			location = varManager.performStringSubstitution(IDFUtil.getIDFPythonEnvPath());
 		}
```
sigmaaa, line 126: theoretically, we shouldn't get there, because it's not allowed to save empty locations now. Mb it's worth deleting this part.

bundles/com.espressif.idf.launch.serial.core/src/com/espressif/idf/launch/serial/internal/SerialFlashLaunchConfigDelegate.java
```diff
@@ -190,7 +192,7 @@ protected void launchInternal(ILaunchConfiguration configuration, String mode, I
 			buffer.append('=').append(entry.getValue());
 			strings.add(buffer.toString());
 		}
-
+		Logger.log(String.format("flash command: %s", String.join(" ", commands))); //$NON-NLS-1$ //$NON-NLS-2$
```
sigmaaa, line 195: can be replaced with the Messages class, but we are already using Messages from org.eclipse.cdt.debug.internal.core package in this class, which btw provides confusing error messages in our case. Maybe it is worth to do refactoring to provide our own Messages class, which will make more sense in our use cases

bundles/com.espressif.idf.launch.serial.core/src/com/espressif/idf/launch/serial/internal/SerialFlashLaunchConfigDelegate.java
```diff
@@ -117,10 +117,12 @@ protected void launchInternal(ILaunchConfiguration configuration, String mode, I
 	{
 		IStringVariableManager varManager = VariablesPlugin.getDefault().getStringVariableManager();
 
-		String location = varManager.performStringSubstitution(IDFUtil.getIDFPythonEnvPath());
+		String location = configuration.getAttribute(ICDTLaunchConfigurationConstants.ATTR_LOCATION,
+				SYSTEM_PATH_PYTHON);
```
alirana01, line 121: Shouldn't the default value come from idf python env path or is it okay to give system path? 🤔

kolipakakondal: Yes, will rely only on IDF_PYTHON_ENV for build and flash

sigmaaa: @kolipakakondal, @alirana01 removed system python path in the latest commit

bundles/com.espressif.idf.launch.serial.core/src/com/espressif/idf/launch/serial/internal/SerialFlashLaunchConfigDelegate.java
kolipakakondal: That make sense, we can create a Message class inside the plugin to avoid internal reference issues

kolipakakondal (approved):
LGTM

kolipakakondal:
Hi @AndriiFilippov could you please verify this?

kolipakakondal (approved):
LGTM

bundles/com.espressif.idf.launch.serial.core/src/com/espressif/idf/launch/serial/internal/SerialFlashLaunchConfigDelegate.java
alirana01: Do we need Messages class for logger? We need to be careful when utilizing the Messages in our Logs as they are language dependent and incase of a different language the logs may switch to that language making it difficult when we try to debug any issue in future with logs.

alirana01 (commented):
LGTM. Just a bit confused on that Messages class comment left a comment there so please take a look at that

AndriiFilippov (approved):

PR #770, opened 2023-05-25
IEP-952: Espressif IDE popup for old tool versions popping up after installing tools via IDE

alirana01:
## Description

Checking the esp_idf.json timestamp for env initialization

Fixes # ([IEP-952](https://jira.espressif.com:8443/browse/IEP-952))

## Type of change

Please delete options that are not relevant.
- Bug fix (non-breaking change which fixes an issue)

## How has this been tested?

Use the tests in the following PR (https://github.com/espressif/idf-eclipse-plugin/pull/662). 
Also try to install the tools in the IDE after the IDE is initialized via esp_idf.json file and try to close and reopen the IDE to see if the message popup appears. The original jira ticket also contains links to the forum and github issues.

**Test Configuration**:
* ESP-IDF Version: Any
* OS (Windows,Linux and macOS): Windows only (Since installer is currently given for Windows only)

## Checklist
- [ ] PR Self Reviewed
- [ ] Applied Code formatting

PR #781, opened 2023-06-07
IEP-974: Adding the IDE env versions to the esp_idf.json

alirana01:
## Description

Previously only the path was changed in the file and now the esp_idf.json file is modified keeping the formatting proper and also adding the object for a different version.

Also improved the message to user

Fixes # ([IEP-974](https://jira.espressif.com:8443/browse/IEP-974))

## Type of change

- Breaking change (fix or feature that would cause existing functionality to not work as expected)
- This change requires a documentation update

## How has this been tested?
Use the testing from IEP-952 ([PR](https://github.com/espressif/idf-eclipse-plugin/pull/770))
Additionally verify that the updated version in the IDE is also added to the esp_idf.json file and is the activated one

**Test Configuration**:
* ESP-IDF Version: Any
* OS (Windows,Linux and macOS): Windows only

## Checklist
- [x] PR Self Reviewed
- [x] Applied Code formatting
- [x] Added Documentation


FAQ.md
alirana01, line 1: Please see if this is fine here or we can somehow add it to main README.MD

bundles/com.espressif.idf.core/src/com/espressif/idf/core/build/messages.properties
alirana01, line 1: Suggestions needed on how we can improve this message

bundles/com.espressif.idf.ui/src/com/espressif/idf/ui/InitializeToolsStartup.java
alirana01, line 1: Needed to use gson from google's api to make the file output pretty.

bundles/com.espressif.idf.core/src/com/espressif/idf/core/build/messages.properties
```diff
@@ -17,5 +17,5 @@ CMakeErrorParser_NotAWorkspaceResource=Could not map %s to a workspace resource.
 IDFBuildConfiguration_CMakeBuildConfiguration_NoToolchainFile=No toolchain file found
 IncreasePartitionSizeTitle=Low Application Partition Size
 IncreasePartitionSizeMessage=Less than 30% of application partition size is free({0} of {1} bytes), would you like to increase it? Please click <a href={2}>here</a> to check more details.
-ToolsInitializationDifferentPathMessageBoxTitle=Different IDF path found in the config file
-ToolsInitializationDifferentPathMessageBoxMessage=A different IDF path was found in the startup configuration file do you want to install the tools in this path or keep the older path\nNew Path: {0}\nOld Path: {1}
+ToolsInitializationDifferentPathMessageBoxTitle=Different IDF path found in IDE Environment
+ToolsInitializationDifferentPathMessageBoxMessage=Different IDF tools are installed in the IDE Environment: {0} than the one selected in startup config file for IDE (esp_idf.json).\nIf you want to keep the IDF tools in the IDE Environment select Yes(This will also add the tools to esp_idf.json).\nIf you want to revert to the last used tools in esp_idf.json press No.
```
sigmaaa, line 21: what about rephrasing it like this:
* 1 version:
The IDE Environment has different IDF tools installed compared to the ones selected in the startup-config file for the IDE (esp_idf.json). 
To keep the IDF tools {0} in the IDE Environment and update the tools in esp_idf.json, select "Yes."
To revert to the last used tools in esp_idf.json, press "No."


However, I found version 1 too verbose
* 2 version:
The IDF tools installed in the IDE Environment are different from the ones selected in the startup-config file (esp_idf.json) for the IDE. Do you want to keep the IDF tools {0} in the IDE Environment and update the tools in esp_idf.json?

what do you think?

bundles/com.espressif.idf.ui/src/com/espressif/idf/ui/InitializeToolsStartup.java
```diff
@@ -250,26 +282,13 @@ private void checkForUpdatedVersion(File idf_json_file)
...
-			}
-			catch (BackingStoreException e)
+			if (list != null)
 			{
-				Logger.log(e);
+				// selected esp-idf version information
+				JSONObject selectedIDFInfo = (JSONObject) list.get(idfVersionId);
+				String idfPath = (String) selectedIDFInfo.get(IDF_PATH);
+				IDFEnvironmentVariables idfEnvMgr = new IDFEnvironmentVariables();
+				return !idfEnvMgr.getEnvValue(IDFEnvironmentVariables.IDF_PATH).equals(idfPath);
```
kolipakakondal, line 291: idfEnvMgr.getEnvValue(IDFEnvironmentVariables.IDF_PATH) will return null when IDE is not configured with the IDF_PATH(IDE is not launched and configured not even once). That leads to NPE

kolipakakondal: In case of multiple back slashes in the path, this might fail.
IDE would have configured IDF_PATH to `C:\esp\esp-idf-5.0` and windows installer would have set to `C:\\esp\\esp-idf-5.0` as per equals() both will be different

bundles/com.espressif.idf.ui/src/com/espressif/idf/ui/InitializeToolsStartup.java
```diff
@@ -102,25 +106,32 @@ public void run()
...
+					String version = IDFUtil.getEspIdfVersion();
+					java.util.regex.Pattern p = java.util.regex.Pattern.compile("([0-9][.][0-9])"); //$NON-NLS-1$
+					java.util.regex.Matcher m = p.matcher(version);
+					if (m.find())
+					{
+						version = m.group(0);
+					}
+					updateEspIdfJsonFile(idf_json_file, idfEnvMgr.getEnvValue(IDFEnvironmentVariables.IDF_PATH),
+							IDFUtil.getIDFPythonEnvPath(),
+							version);
```
kolipakakondal, line 134: What if the user updated esp-idf(let's say 5.0 to 5.1) through windows installer, in this case `isVersionDifferentInFile` is `true` considering we compare esp-idf path from IDE and esp-idf.json

Don't we export the paths in this case?

kolipakakondal (changes requested):
Probably, we can also update message box with more clear and concise message

bundles/com.espressif.idf.ui/src/com/espressif/idf/ui/InitializeToolsStartup.java
sigmaaa: good catch. Probably we could use Paths.get().normalize() here

bundles/com.espressif.idf.core/src/com/espressif/idf/core/build/messages.properties
alirana01: The second version is misleading to my understanding as I think it can confuse the user like they can press No thinking that the tools wont be added to json file but instead we are trying to revert the tools back to the original installation. This would be okay if we didnt have the usecase where the tools are updated via the IDE offline installer.

bundles/com.espressif.idf.ui/src/com/espressif/idf/ui/InitializeToolsStartup.java
alirana01: > good catch. Probably we could use Paths.get().normalize() here

I dont think using normalize is a good idea simply comparison like using `Paths.get(path).toAbsolutePath().compareTo(pathToCompare) == 0`
should be enough for us considering that we null check first.

alirana01:
Closing this PR as this is not anything we need because of EIM introduction

PR #784, opened 2023-06-09
IEP-976 Test coverage for BigIntDecoder

sigmaaa:
## Description

Added test coverage for BigIntDecoder class. Also, during the test I found bug, so I fixed it in this PR. The bug was in the incorrect decoding of hex, and octal values without a sign.

Fixes # ([IEP-976](https://jira.espressif.com:8443/browse/IEP-976))

## Type of change

## How has this been tested?
BigIntDecoder is using for validating data in the NVS table, so to test it:
- Open the NVS table editor
- Add a new row 
- type some name in a new row and select 'data' as a type 
- select encoding, for example u8 and type hex/oct/dec numbers for example 0xFF - should be ok, 0xFFF - out of range for u8, FF - incorrect format

**Test Configuration**:
* ESP-IDF Version:

- NVS table editor validation input

## Checklist
- [x] PR Self Reviewed
- [x] Applied Code formatting
- [ ] Added Documentation
- [x] Added Unit Test

### DIFF
--- a/bundles/com.espressif.idf.core/src/com/espressif/idf/core/util/BigIntDecoder.java
+++ b/bundles/com.espressif.idf.core/src/com/espressif/idf/core/util/BigIntDecoder.java
@@ -45,7 +45,7 @@ public class BigIntDecoder
 		}
 
 		return hasSign ? new BigInteger(firstChar + stringToDecode.substring(index), radix)
-				: new BigInteger(stringToDecode.substring(index));
+				: new BigInteger(stringToDecode.substring(index), radix);
 	}
 
 }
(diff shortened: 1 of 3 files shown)


bundles/com.espressif.idf.core/src/com/espressif/idf/core/util/BigIntDecoder.java
```diff
@@ -45,7 +45,7 @@ else if (stringToDecode.startsWith("0", index) && stringToDecode.length() > 1 +
 		}
 
 		return hasSign ? new BigInteger(firstChar + stringToDecode.substring(index), radix)
-				: new BigInteger(stringToDecode.substring(index));
+				: new BigInteger(stringToDecode.substring(index), radix);
```
sigmaaa, line 48: the bug was here. Without radix, we can't parse hex, oct values

sigmaaa (commented):
self reviewed

kolipakakondal (commented):
LGTM

PR #788, opened 2023-06-19
IEP-984 org.slf4j.api dependency cannot be resolved

sigmaaa:
looks like org.slf4j.api was removed from the egit dependencies


alirana01 (approved):
LGTM!

PR #865, opened 2023-12-08
IEP-1121: build/ide/idf_components: Avoid showing error markers on the project

alirana01:
## Description

The ide/esp_idf_components contained derived files and due to cdt indexer issues they showed invalid errors and which was not good UX. The following PR creates a Listener for the markers that are added in the problems view. A preference checkbox is also added to enable or disable this
image

Fixes # ([IEP-1121](https://jira.espressif.com:8443/browse/IEP-1121))

## Type of change

## How has this been tested?

Build a project and then go into build/ide/esp_idf_components and try to open the source files from the components their must not be any errors in that 
@AndriiFilippov also test by enabling and disabling the added setting

**Test Configuration**:
* ESP-IDF Version: any
* OS (Windows,Linux and macOS): all

## Checklist
- [x] PR Self Reviewed
- [x] Applied Code formatting

## Summary by CodeRabbit

- **New Features**
  - Added a class `ErrorMarkerListener` to listen for added markers in the problems view and trigger a job to remove them from the derived files in the IDE folder for `esp_idf` components.
  - Introduced a marker cleanup job triggered when relevant files are opened to improve the user experience by removing unwanted errors shown by the CDT indexer.


alirana01 (commented):
Self Review

bundles/com.espressif.idf.core/src/com/espressif/idf/core/build/EditorOpenListener.java
```diff
@@ -0,0 +1,50 @@
...
+		{
+			if (partRef instanceof IEditorReference)
+			{
+				IEditorReference editorRef = (IEditorReference) partRef;
+				if (editorRef.getEditorInput() instanceof IEditorInput)
+				{
+					IEditorInput fileInput = editorRef.getEditorInput();
+					IFile file = fileInput.getAdapter(IFile.class);
+					String filePath = file.getFullPath().toString();
+					if (filePath.contains("/ide/esp_idf_components/")) //$NON-NLS-1$
```
sigmaaa, line 36: is it possible to have the path separated by `\` here?

sigmaaa (commented):
LGTM

bundles/com.espressif.idf.core/src/com/espressif/idf/core/build/EditorOpenListener.java
alirana01: I think the getFullPath will return the path with `/` instead of `\` for all operating systems so its better to use that instead of `\`

kolipakakondal:
Hi @alirana01 Can you please check this, I have opened flash_mmap.c file from the spi_flash component and the below errors reported on the project and problems view.

Screenshot 2023-12-26 at 1 39 14 PM

kolipakakondal:
Here is one more example.
Screenshot 2023-12-26 at 1 45 31 PM

AndriiFilippov:
@alirana01 hi !

LGTM 👍

kolipakakondal:
Thanks @alirana01 for quickly implementing this!

AndriiFilippov:
@alirana01 hi !

when the option is disabled - it seams to be working as enabled caz I still can't see errors 
image

bundles/com.espressif.idf.core/src/com/espressif/idf/core/build/ErrorMarkerListener.java
```diff
@@ -0,0 +1,104 @@
...
+ * @author Ali Azam Rana
+ *
+ */
+public class ErrorMarkerListener implements IResourceChangeListener
+{
+	private static String IDF_PATH;
+
+	public ErrorMarkerListener()
+	{
+		IDF_PATH = new IDFEnvironmentVariables().getEnvValue(IDFEnvironmentVariables.IDF_PATH);
```
sigmaaa, line 36: Do we need to initialize the IDF_PATH field in the constructor? As far as I understand, we assign this listener when starting the IDE. In this case, the IDF_PATH will be null on the first startup, when the environment variables have not yet been initialized. Also, this variable will not be changed after the tools are installed.

bundles/com.espressif.idf.core/src/com/espressif/idf/core/build/ErrorMarkerListener.java
```diff
@@ -0,0 +1,104 @@
...
+			try
+			{
+				if (!(delta.getMarker().getResource() instanceof IFile))
+				{
+					return;
+				}
+
+				IFile file = (IFile) delta.getMarker().getResource();
+				if (file.exists() && delta.getKind() == IResourceChangeEvent.POST_CHANGE
+						&& delta.getMarker().isSubtypeOf(IMarker.PROBLEM) && Integer.valueOf(
```
sigmaaa, line 60: Better use Integer.parseInt here to avoid unboxing

sigmaaa (commented):
I think there is a small problem with the IDF_PATH assignment in the listener constructor, otherwise LGTM!

bundles/com.espressif.idf.core/src/com/espressif/idf/core/build/ErrorMarkerListener.java
alirana01: Fixed

bundles/com.espressif.idf.core/src/com/espressif/idf/core/build/ErrorMarkerListener.java
alirana01: Done

AndriiFilippov (approved):

Issue #1105, opened 2024-12-28
Unable to install ESP-IDF 5.3.2 via ESP-IDF Manager in Elclipse (IEP-1395)


sim0njo:
**Describe the bug**
Unable to install ESP-IDF from ESP-IDF Manager, tried downloading it, tried using a downloaded version, tried it 10 times
This really annoyes me
Your documentation about installing the ESP-IDF in Eclipse is also like going in circles, missing concise and working procedure !!!

**To Reproduce**
1. Go to 'ESP-IDF manager' in Eclipse
2. Click on 'Add ESP-IDF' -> download ESP-IDF to C:\Espressif -> press Finish
>>>
Selected targets are: esp32c6, esp32h2, esp32s3, esp32c61, esp32c2, esp32c3, esp32c5, esp32, esp32p4, esp32s2
Current system platform: win64
Installing tools: xtensa-esp-elf-gdb, riscv32-esp-elf-gdb, xtensa-esp-elf, esp-clang, riscv32-esp-elf, esp32ulp-elf, cmake, openocd-esp32, ninja, idf-exe, ccache, dfu-util, esp-rom-elfs, qemu-xtensa, qemu-riscv32
Skipping xtensa-esp-elf-gdb@14.2_20240403 (already installed)
Skipping riscv32-esp-elf-gdb@14.2_20240403 (already installed)
Skipping xtensa-esp-elf@esp-13.2.0_20240530 (already installed)
Skipping esp-clang@16.0.1-fe4f10a809 (already installed)
Skipping riscv32-esp-elf@esp-13.2.0_20240530 (already installed)
Skipping esp32ulp-elf@2.38_20240113 (already installed)
Skipping cmake@3.30.2 (already installed)
Skipping openocd-esp32@v0.12.0-esp32-20241016 (already installed)
Skipping ninja@1.12.1 (already installed)
Skipping idf-exe@1.0.3 (already installed)
Skipping ccache@4.10.2 (already installed)
Skipping dfu-util@0.11 (already installed)
Skipping esp-rom-elfs@20240305 (already installed)
Skipping qemu-xtensa@esp_develop_9.0.0_20240606 (already installed)
Skipping qemu-riscv32@esp_develop_9.0.0_20240606 (already installed)

Executing C:\Users\simonjo\AppData\Local\Microsoft\WindowsApps\python3.EXE C:\Espressif\esp-idf-v5.3.2\tools\idf_tools.py install-python-env
Python 3.11.2
[...]

PR #1110, opened 2025-01-08
IEP-1395: Removing system env vars from IDE while executing scripts

alirana01:
## Description

The original bug reported here https://github.com/espressif/idf-eclipse-plugin/issues/1105

Fixes # ([IEP-1395](https://jira.espressif.com:8443/browse/IEP-1395))

## Type of change

## How has this been tested?
Please try to put the system env vars in windows as mentioned on original reported issues and then try to install tools to test if they are getting installed with correct path in the IDE.

**Test Configuration**:
* ESP-IDF Version: any
* OS (Windows,Linux and macOS): all

## Checklist
- [x] PR Self Reviewed
- [x] Applied Code formatting

## Summary by CodeRabbit

- **Logging Improvements**
	- Enhanced logging in tools installation process to capture more detailed status messages.

- **Environment Configuration**
	- Refined environment variable management for IDF tools by removing unnecessary entries.
	- Updated handling of Git executable path in command execution environment.

### DIFF
--- a/bundles/com.espressif.idf.ui/src/com/espressif/idf/ui/tools/ToolsJob.java
+++ b/bundles/com.espressif.idf.ui/src/com/espressif/idf/ui/tools/ToolsJob.java
@@ -349,6 +349,11 @@ public abstract class ToolsJob extends Job
 			Logger.log(environment.toString());
 			environment.put("PYTHONUNBUFFERED", "1"); //$NON-NLS-1$ //$NON-NLS-2$
 			loadIdfPathWithSystemPath(environment);
+			environment.remove(IDFEnvironmentVariables.IDF_PYTHON_ENV_PATH);
+			environment.remove(IDFEnvironmentVariables.ESP_IDF_VERSION);
+			environment.remove(IDFEnvironmentVariables.IDF_PATH);
+			environment.remove(IDFEnvironmentVariables.OPENOCD_SCRIPTS);
+			
 			if (gitExecutablePath != null)
 			{
 				addPathToEnvironmentPath(environment, gitExecutablePath);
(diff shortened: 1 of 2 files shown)


bundles/com.espressif.idf.ui/src/com/espressif/idf/ui/tools/ToolsJob.java
```diff
@@ -349,6 +349,11 @@ protected IStatus runCommandIdfPyInIdfEnv(List<String> arguments, MessageConsole
 			Logger.log(environment.toString());
 			environment.put("PYTHONUNBUFFERED", "1"); //$NON-NLS-1$ //$NON-NLS-2$
 			loadIdfPathWithSystemPath(environment);
+			environment.remove(IDFEnvironmentVariables.IDF_PYTHON_ENV_PATH);
+			environment.remove(IDFEnvironmentVariables.ESP_IDF_VERSION);
+			environment.remove(IDFEnvironmentVariables.IDF_PATH);
+			environment.remove(IDFEnvironmentVariables.OPENOCD_SCRIPTS);
```
sigmaaa, line 355: Hi @alirana01, 

Do you know why these variables are not replaced with the same ones from the idfToolSet (which, as I understand, should have proper environment variables)?

alirana01: So basically the environment here is taken from `System.getEnv()` so it loads the system env. The function where the change was done is actually before these variables are even part of the idfToolSet. This function make sures that the correct variables from the idf export script are loaded into the idfToolSet basically it sits before that. Which is the reason we need to make sure that any of these variables from System environment are not loaded here

kolipakakondal (approved):

PR #1170, opened 2025-03-04
IEP-1470: Fix for exception on invalid file and editor for file open listener

alirana01:
## Description

Fix for exception on invalid file and editor for file open listener

Fixes # ([IEP-1470](https://jira.espressif.com:8443/browse/IEP-1470))

## Type of change

## How has this been tested?

Read the original bug description on Jira and try to reproduce no exception in errors of similar nature should be visible.

**Test Configuration**:
* ESP-IDF Version: all
* OS (Windows,Linux and macOS): any

## Checklist
- [x] PR Self Reviewed
- [x] Applied Code formatting

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved file handling to skip processing for non-C source/header files when a dedicated editor view is active, enhancing the file opening experience and preventing potential conflicts during file interactions.

### DIFF
--- a/bundles/com.espressif.idf.ui/src/com/espressif/idf/ui/handlers/FileOpenListener.java
+++ b/bundles/com.espressif.idf.ui/src/com/espressif/idf/ui/handlers/FileOpenListener.java
@@ -42,21 +42,9 @@ public class FileOpenListener implements IPartListener2
 			if (project == null)
 				return;
 			
-			try
+			if (!isSourceOrHeaderFile(file))
 			{
-				IContentTypeManager contentTypeManager = Platform.getContentTypeManager();
-				IContentType contentType = contentTypeManager.findContentTypeFor(file.getFullPath().toOSString());
-				if (contentType != null)
-				{
-					String id = contentType.getId();
-					if (!(id.startsWith("org.eclipse.cdt.core.c") && (id.endsWith("Source") || id.endsWith("Header")))) {
-	                    return; // Skip files that are not C source/header files
-	                }
-				}
-			}
-			catch (Exception e)
-			{
-				Logger.log(e);
+				return;
 			}
 			
 			String buildDir = StringUtil.EMPTY;
@@ -75,6 +63,27 @@ public class FileOpenListener implements IPartListener2
 			lspService.restartLspServers();
 
 		}
+	}
+	
+	private boolean isSourceOrHeaderFile(IFile file)
+	{
+		try
+		{
+			IContentTypeManager contentTypeManager = Platform.getContentTypeManager();
+			IContentType contentType = contentTypeManager.findContentTypeFor(file.getFullPath().toOSString());
+			if (contentType != null)
+			{
+				String id = contentType.getId();
+				if ((id.startsWith("org.eclipse.cdt.core.c") && (id.endsWith("Source") || id.endsWith("Header")))) {
+                    return true;
+                }
+			}
+		}
+		catch (Exception e)
+		{
+			Logger.log(e);
+		}
+		return false;
 	}
 
 	@Override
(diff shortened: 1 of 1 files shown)


alirana01 (commented):
Self Review

bundles/com.espressif.idf.ui/src/com/espressif/idf/ui/handlers/FileOpenListener.java
```diff
@@ -34,6 +35,10 @@ public void partActivated(IWorkbenchPartReference partRef)
 		if (part instanceof IEditorPart ieditorpart
 				&& ieditorpart.getEditorInput() instanceof FileEditorInput fileInput)
 		{
+			
+			if (ieditorpart instanceof ESPIDFManagerEditor)
+				return;
```
sigmaaa, line 40: We can make it more robust because, theoretically, the same exception could occur in another editor. What if we get the content description from the file (fileInput.getFile()) and then compare the content type ID with the following condition: id.startsWith("org.eclipse.cdt.core.c") && (id.endsWith("Source") || id.endsWith("Header"))?

sigmaaa (commented):
Hi @alirana01, LGTM, but I think we can make it more robust. Please check the comment

bundles/com.espressif.idf.ui/src/com/espressif/idf/ui/handlers/FileOpenListener.java
alirana01: That is actually a very nice idea let me try to update this :)

bundles/com.espressif.idf.ui/src/com/espressif/idf/ui/handlers/FileOpenListener.java
```diff
@@ -38,6 +41,24 @@ public void partActivated(IWorkbenchPartReference partRef)
...
+					String id = contentType.getId();
+					if (!(id.startsWith("org.eclipse.cdt.core.c") && (id.endsWith("Source") || id.endsWith("Header")))) {
+	                    return; // Skip files that are not C source/header files
+	                }
+				}
+			}
+			catch (Exception e)
+			{
+				Logger.log(e);
+			}
```
sigmaaa, line 60: LGTM!

Nit: Consider extracting this logic into a private method for better readability. You could then add a single condition to skip processing, like this: 
```
			if (project == null || !isCSourceOrHeader(file))
				return;
```
```
	private boolean isCSourceOrHeader(IFile file)
	{
		try
		{
			IContentTypeManager contentTypeManager = Platform.getContentTypeManager();
			IContentType contentType = contentTypeManager.findContentTypeFor(file.getFullPath().toOSString());
			if (contentType != null)
			{
				String id = contentType.getId();
				return id.startsWith("org.eclipse.cdt.core.c") && (id.endsWith("Source") || id.endsWith("Header")); //$NON-NLS-1$ //$NON-NLS-2$ //$NON-NLS-3$
			}
		}
		catch (Exception e)
		{
			Logger.log(e);
		}
		return false;
	}
```

sigmaaa (approved):
Hi @alirana01, LGTM! Just one small nit-pick and a reminder that NON-NLS annotations are missing

AndriiFilippov:
@alirana01 hi !

OS - Mac arm64
issue still persist

alirana01:
> @alirana01 hi !
> 
> OS - Mac arm64
> issue still persists

Please share error logs

AndriiFilippov:
[LOGS.log](https://github.com/user-attachments/files/19126240/LOGS.log)


@alirana01

alirana01:
> [LOGS.log](https://github.com/user-attachments/files/19126240/LOGS.log)
> 
> @alirana01

@AndriiFilippov 
I couldn't reproduce the same error that this issue was originally reported for there is another exception which is unrelated and we can proceed without that. Just a side note kindly clear the logs before you execute a flow to test when sharing logs

sigmaaa (approved):

PR #1171, opened 2025-03-05
IEP-1467 Update clangd path based on the esp-idf and tools

sigmaaa:
## Description

Updating the clangd path after tools activation and shutting down the server before the restart to avoid this issue:
image

Fixes # ([IEP-1467](https://jira.espressif.com:8443/browse/IEP-1467))

## Type of change

## How has this been tested?

1) install ESP-IDF (master in my case) and simply create a project - then open main.c - no errors and function definition and auto-completion are available 
2) Install multiple esp-idf versions and change between them -> check clangd path is updated accordingly 

**Test Configuration**:
* ESP-IDF Version:

- LSP server

## Checklist
- [ ] PR Self Reviewed
- [ ] Applied Code formatting
- [ ] Added Documentation
- [ ] Added Unit Test

## Summary by CodeRabbit

- **New Features**
  - Introduced an update that automatically adjusts language service configuration paths during tool activation.
  - Added a user-visible message indicating when the language server is restarting.

- **Refactor**
  - Streamlined the language server restart workflow for improved reliability.
  - Improved message formatting for clearer display.

- **Chores**
  - Updated package dependencies to support enhanced language service preferences.

### DIFF
--- a/bundles/com.espressif.idf.core/src/com/espressif/idf/core/util/LspService.java
+++ b/bundles/com.espressif.idf.core/src/com/espressif/idf/core/util/LspService.java
@@ -16,8 +16,6 @@ import org.eclipse.lsp4e.LanguageServerWrapper;
 import org.eclipse.lsp4e.LanguageServiceAccessor;
 import org.eclipse.ui.PlatformUI;
 
-import com.espressif.idf.core.logging.Logger;
-
 @SuppressWarnings("restriction")
 public class LspService
 {
@@ -29,6 +27,7 @@ public class LspService
 		this(PlatformUI.getWorkbench().getService(ClangdConfiguration.class),
 				LanguageServiceAccessor.getStartedWrappers(null, true).stream()
 						.filter(w -> "org.eclipse.cdt.lsp.server".equals(w.serverDefinition.id)).toList()); //$NON-NLS-1$
+
 	}
 
 	public LspService(Configuration configuration, List<LanguageServerWrapper> languageServerWrappers)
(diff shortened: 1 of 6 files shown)


kolipakakondal (approved):

alirana01 (approved):
LGTM

alirana01:
Generally the PR is fine but on another note I wanted to discuss some idea on how to stream line these calls to different services.
I see that LspService is used at multiple places but considering we are calling this a service shouldn't there be only one instance of the service handling all these calls? 
Can we use some kind of listener to make these calls only in one place based on some filters for what is calling that listener?

I think we can probably create a ticket to discuss this a bit more and see what we can come up with.

For the scope of this ticket it is fine and good to go

sigmaaa:
> Generally the PR is fine but on another note I wanted to discuss some idea on how to stream line these calls to different services.
I see that LspService is used at multiple places but considering we are calling this a service shouldn't there be only one instance of the service handling all these calls?

Hi @alirana01, thanks for the review! I considered using a Singleton here, but since the operations in the LspService constructor are not costly, I believe it's not necessary in this case. That said, we can definitely create a separate ticket to discuss this further if needed

AndriiFilippov:
@sigmaaa hi !

Tested under: 
OS - Mac arm64 

install ESP-IDF -create a project - then open main.c - no errors and function definition and auto-completion are available ✅ 
install ESP-IDF -create a project - build flash monitor - then open main.c - no errors and function definition and auto-completion are available ✅ 
Install multiple esp-idf versions and change between them -> check clangd path is updated accordingly ✅

PR #1217, opened 2025-05-07
[WIP] Release/v4.0.0 This is just to have the release 4.0 builds for testing (Please dont merge it yet

alirana01:


bundles/com.espressif.idf.serial.monitor/src/com/espressif/idf/serial/monitor/core/IDFMonitor.java
```diff
@@ -162,7 +155,7 @@ public Process start() throws Exception
 		idfEnvMap.put("PYTHONUNBUFFERED", "1"); //$NON-NLS-1$ //$NON-NLS-2$
 
 		// Update with the CDT build environment variables
-		Map<String, String> environment = new HashMap<>(IDFUtil.getSystemEnv());
+		Map<String, String> environment = new HashMap<>(System.getenv());
```
sigmaaa, line 158: We should verify whether this is still necessary. If the ESP-IDF environment variables are already present in the system environment, we likely don’t need to pass them explicitly — the local terminal should inherit them automatically. The same applies to the IDF terminal.

sigmaaa (commented):
Hi @alirana01,

Thanks for the PR! It looks good overall. I’ve left a few comments with suggestions we can check together.

PR #1267, opened 2025-07-09
IEP-1596 Handle Redundant File Watcher Events for eim_idf.json

sigmaaa:
## Description

Added a 100ms delay before processing file system events to prevent handling multiple ENTRY_MODIFY events triggered by certain text editors. This helps avoid redundant processing when eim_idf.json is edited rapidly.

I saw this behavior a few times during the migration from 3.5 to 4.0.0, without manually editing the eim_idf.json file. I couldn't reproduce it since then, so I tried editing the file manually and observed similar behavior.

Fixes # ([IEP-1596](https://jira.espressif.com:8443/browse/IEP-1596))

## Type of change

## How has this been tested?

open the IDE -> edit esp_idf.json file -> should be one notification in the IDE

**Test Configuration**:
* ESP-IDF Version:

- Component 1
- Component 2

## Checklist
- [ ] PR Self Reviewed
- [ ] Applied Code formatting
- [ ] Added Documentation
- [ ] Added Unit Test

### DIFF
--- a/bundles/com.espressif.idf.core/src/com/espressif/idf/core/tools/watcher/EimJsonWatchService.java
+++ b/bundles/com.espressif.idf.core/src/com/espressif/idf/core/tools/watcher/EimJsonWatchService.java
@@ -5,8 +5,15 @@
 package com.espressif.idf.core.tools.watcher;
 
 import java.io.IOException;
-import java.nio.file.*;
-import java.nio.file.attribute.FileAttribute;
+import java.nio.file.ClosedWatchServiceException;
+import java.nio.file.FileSystems;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.nio.file.StandardWatchEventKinds;
+import java.nio.file.WatchEvent;
+import java.nio.file.WatchKey;
+import java.nio.file.WatchService;
 import java.util.List;
 import java.util.concurrent.CopyOnWriteArrayList;
 
@@ -28,8 +35,7 @@ public class EimJsonWatchService extends Thread
 	private final List<EimJsonChangeListener> eimJsonChangeListeners = new CopyOnWriteArrayList<>();
 	private volatile boolean running = true;
 	private volatile boolean paused = false;
-	
-	
+
 	private EimJsonWatchService() throws IOException
 	{
 		String directoryPathString = Platform.getOS().equals(Platform.OS_WIN32) ? EimConstants.EIM_WIN_DIR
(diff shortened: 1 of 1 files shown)


bundles/com.espressif.idf.core/src/com/espressif/idf/core/tools/watcher/EimJsonWatchService.java
```diff
@@ -139,6 +145,17 @@ public void run()
...
+			{
+				// Prevent handling multiple ENTRY_MODIFY events (e.g., when eim_idf.json is edited in a text editor
+				// like VS Code)
+				Thread.sleep(100);
+			}
+			catch (InterruptedException e)
+			{
+				Logger.log(e);
+				Thread.currentThread().interrupt();
+			}
```
kolipakakondal: Can we address this with the File lastmodifiedtime? Save the last modified time in the memory and during every event check it was really modified by comparing the last modified timestamp..does that work or content hash approach?

d-almazov: Hi @kolipakakondal, I think your suggestion would work, but it feels like overkill for this case, and it would require more resources (e.g., additional memory allocation, extra conditional checks).
From what I’ve seen, this issue occurs because editing the file in some editors—or possibly through eim.exe—triggers multiple ENTRY_MODIFY events at once. As a result, we show a notification for each event.
A small delay in this thread helps group those events together, so we show only a single notification to update the Tools Manager.

kolipakakondal: That’s my concern - we don’t know the exact root cause, and we’re trying to avoid the issue by adding a time delay. But I’m not sure that will work 100%.

d-almazov: Hi @kolipakakondal, the root cause is that a file modification can trigger multiple ENTRY_MODIFY events—for example, one for updating the content and another for updating file metadata. In such cases, a small delay (e.g., Thread.sleep(100)) is enough to debounce the duplicate events. I believe a similar issue is discussed here: https://stackoverflow.com/questions/16777869/java-7-watchservice-ignoring-multiple-occurrences-of-the-same-event

d-almazov: Hi @kolipakakondal,

In the last commit, I replaced the debounce solution with a comparison of file timestamps.

sigmaaa:
In the last commit, I replaced the debounce solution with comparing file timestamps, as suggested by Kondal.

AndriiFilippov:
@sigmaaa hi ! 

Tested under:
OS - Windows 11 / Mac arm64

redundant notifications issue fixed ✅ 
LGTM 👍

AndriiFilippov (approved):

PR #1269, opened 2025-07-10
fix: Add python path for build commands

kolipakakondal:
## Description

Add python path for build commands to avoid build issues.

Fixes # ([IEP-1580](https://jira.espressif.com:8443/browse/IEP-1580))

## Type of change

## How has this been tested?

- Create a project
- Build the project and verify that CMake build command is appended with the idf python path

**Test Configuration**:
* ESP-IDF Version: master
* OS (Windows,Linux and macOS):macOS

- Build

## Checklist
- [ ] PR Self Reviewed
- [ ] Applied Code formatting
- [ ] Added Documentation
- [ ] Added Unit Test


bundles/com.espressif.idf.core/src/com/espressif/idf/core/build/IDFBuildConfiguration.java
```diff
@@ -431,8 +431,13 @@ private void runCmakeCommand(IConsole console, IProgressMonitor monitor, IProjec
...
 		infoStream.write(String.format(Messages.CMakeBuildConfiguration_Configuring, buildDir));
 
+		IDFEnvironmentVariables envVariables = new IDFEnvironmentVariables();
+		String pythonPath = envVariables.getEnvValue(IDFEnvironmentVariables.IDF_PYTHON_ENV_PATH);
 		List<String> command = new ArrayList<>();
-
+		if (!StringUtil.isEmpty(pythonPath))
+		{
+			command.add(pythonPath);
+		}
```
d-almazov: Hi @kolipakakondal. Is this correct? Would this run CMake as a Python script?
I think the proper way is to call cmake directly and pass the Python path using the -DPYTHON=... argument instead.

Also, do we need to include the following arguments as well?
```
command.add("-DPYTHON_DEPS_CHECKED=1");  // Skip Python dependency checks
command.add("-DPYTHON=" + IDFUtil.getIDFPythonEnvPath()); // Use specific Python path
command.add("-DESP_PLATFORM=1"); // Enable ESP-IDF platform-specific behavior
```

sigmaaa (commented):
Hi @kolipakakondal, I'm not sure if that's the correct way to specify Python in the CMake build command. I’ve left a comment.

kolipakakondal:
Hi @sigmaaa Thanks for reviewing. Addressed the PR comments.

PR #1271, opened 2025-07-14
IEP-1600 EspConfigParser refactoring

sigmaaa:
## Description

The goal of this PR is to optimize EspConfigParser by using faster gson instead of json-simple, load the config file only once in the constructor, and avoid unnecessary exceptions like java.io.FileNotFoundException: \esp-config.json (The system cannot find the file specified)

Fixes # ([IEP-1600](https://jira.espressif.com:8443/browse/IEP-1600))

## Type of change

## How has this been tested?

- Verify Launch Target by comparing the selected target and the boards/config that appear
- Verify Launch/Flash/Debug
- Verify there is no exception when esp-config.json is missing

**Test Configuration**:
* ESP-IDF Version:

- Launch Target
- New Project Wizard
- Launch/Debug

## Checklist
- [ ] PR Self Reviewed
- [ ] Applied Code formatting
- [ ] Added Documentation
- [ ] Added Unit Test

## Summary by CodeRabbit

* **Refactor**
  * Improved performance and reliability of ESP configuration handling by optimizing how configuration files are read and parsed.
  * Enhanced type safety and error handling for ESP board and target configurations.
  * Updated UI components to use the new board data model for displaying and selecting ESP boards.

* **Tests**
  * Added comprehensive tests to verify correct parsing and retrieval of ESP targets, flash voltages, and board configurations.

* **Chores**
  * Introduced a new sample configuration file for testing and reference.
  * Updated package exports to support the new configuration parser structure.


AndriiFilippov (approved):

bundles/com.espressif.idf.core/src/com/espressif/idf/core/util/EspConfigParser.java
```diff
@@ -15,124 +14,126 @@
...
+	}
+
+	private static class Option
+	{
+		String name;
+		List<String> values;
+	}
+
+	private static class Board
+	{
```
alirana01: Hi @sigmaaa this doesn't feel right to have these classes as static inside the parser. Can we please move them to a separate package. I guess the whole class should be moved out of util and inside a package lets say named configparser 

configparser 
|----- vo
            | --- EspConfig.java
            | --- Target
            | --- Option
            | --- ***
|----- EspConfigParser.java


This will keep extensibility and SRD intact and if you think we need to avoid reuse why not just keep the visibility to default or package only?

bundles/com.espressif.idf.launch.serial.ui/src/com/espressif/idf/launch/serial/ui/internal/NewSerialFlashTargetWizardPage.java
alirana01, line 1: If we take care of moving POJOs out of the class as static I guess we can use them directly here and avoid having a map.

bundles/com.espressif.idf.core/src/com/espressif/idf/core/util/EspConfigParser.java
d-almazov: Hi @alirana01,
Thanks a lot for the feedback!

I see your point about SRP and extensibility. I kept the models as static inner classes since they’re small, tightly scoped to the parser, and not reused elsewhere. The goal was to keep parsing logic and related structures encapsulated.

From an SRP perspective, I think this still holds, and I followed the YAGNI principle here, as I don’t expect these models to grow or be reused.

That said, I fully agree that using POJOs over maps improves clarity and type safety — and in this case, public inner classes could serve that purpose too.

Your suggestion to move them to separate files also makes sense — just wanted to share the reasoning behind the current setup.

sigmaaa:
Hi @alirana01,

Thanks for the review. I've addressed the comments in the commit.

AndriiFilippov:
LGTM 👍

AndriiFilippov (approved):

PR #1274, opened 2025-07-18
fix: write openocd debug output to a file

kolipakakondal:
## Description

This PR adds the capability to capture and write OpenOCD debug output to a file, providing better debugging and logging capabilities 

Fixes # ([IEP-1536](https://jira.espressif.com:8443/browse/IEP-1536))

## Type of change

## How has this been tested?
- Open Debug configuraiton
- Navigate to common tab and choose output file
- Debug the application
- Verify that output is written to a selected file

Scenarios verified

- Tested with existing directory paths
- Tested with non-existent directory paths (auto-creation)
- Tested with explicit file paths
- Tested append 

**Test Configuration**:
* ESP-IDF Version: master
* OS (Windows,Linux and macOS): macOS

- Debugging

## Checklist
- [ ] PR Self Reviewed
- [ ] Applied Code formatting
- [ ] Added Documentation
- [ ] Added Unit Test

### DIFF
--- a/bundles/com.espressif.idf.debug.gdbjtag.openocd/src/com/espressif/idf/debug/gdbjtag/openocd/dsf/process/StreamListener.java
+++ b/bundles/com.espressif.idf.debug.gdbjtag.openocd/src/com/espressif/idf/debug/gdbjtag/openocd/dsf/process/StreamListener.java
@@ -6,11 +6,14 @@ package com.espressif.idf.debug.gdbjtag.openocd.dsf.process;
 
 import java.io.BufferedReader;
 import java.io.File;
+import java.io.FileWriter;
 import java.io.IOException;
+import java.io.PrintWriter;
 import java.io.StringReader;
 import java.nio.charset.Charset;
 import java.util.List;
 
+import org.eclipse.core.variables.VariablesPlugin;
 import org.eclipse.debug.core.IStreamListener;
 import org.eclipse.debug.core.model.IFlushableStreamMonitor;
 import org.eclipse.debug.core.model.IProcess;
@@ -50,11 +53,24 @@ public class StreamListener implements IStreamListener
 	private boolean fStreamClosed = false;
 	private List<ReHintPair> reHintsList;
 
+	private String outputFile;
+	private PrintWriter fileWriter;
+
 	public StreamListener(IProcess iProcess, IStreamMonitor errorStreamMonitor, IStreamMonitor outputStreamMonitor,
-			Charset charset)
+			Charset charset, String outputFile, boolean append)
 	{
 		fErrorStreamMonitor = errorStreamMonitor;
 		fOutputStreamMonitor = outputStreamMonitor;
+		this.outputFile = outputFile;
+		String resolvedOutputFile = resolveOutputFilePath(outputFile);
+		if (resolvedOutputFile != null && !resolvedOutputFile.isEmpty()) {
+			try {
+				fileWriter = new PrintWriter(new FileWriter(resolvedOutputFile, append), true);
+			} catch (IOException e) {
+				Logger.log(e);
+				fileWriter = null;
+			}
+		}
 
 		idfProcessConsole = IdfProcessConsoleFactory.showAndActivateConsole(charset);
 		// Clear the console only at the beginning, when OpenOCD starts
@@ -112,6 +128,9 @@ public class StreamListener implements IStreamListener
 		{
 			while ((line = bufferedReader.readLine()) != null)
 			{
+				if (outputFile != null && !outputFile.isEmpty() && fileWriter != null) {
+					fileWriter.println(line);
+				}
 
 				if (line.startsWith("Error:") && fConsoleErrorOutputStream != null) //$NON-NLS-1$
 				{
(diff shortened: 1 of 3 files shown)


bundles/com.espressif.idf.debug.gdbjtag.openocd/src/com/espressif/idf/debug/gdbjtag/openocd/dsf/process/StreamListener.java
```diff
@@ -186,5 +205,38 @@ public void dispose()
...
+				parent.mkdirs();
+			}
+			if (!file.exists()) {
+				file.createNewFile();
+			}
+			return file.getAbsolutePath();
+		} catch (Exception e) {
+			Logger.log(e);
+			return null;
+		}
```
d-almazov: It seems the formatter isn't being applied here.

bundles/com.espressif.idf.debug.gdbjtag.openocd/src/com/espressif/idf/debug/gdbjtag/openocd/dsf/process/StreamListener.java
```diff
@@ -112,6 +128,9 @@ public void streamAppended(String text, IStreamMonitor monitor)
 		{
 			while ((line = bufferedReader.readLine()) != null)
 			{
+				if (outputFile != null && !outputFile.isEmpty() && fileWriter != null) {
```
d-almazov: NIT: We could avoid the conditional check by using Writer.nullWriter() instead. Also, consider using an autoclosable and defining fileWriter locally here, rather than as a field, as suggested in another comment.
For example, the beginning of this method could look like this:

```
	@Override
	public void streamAppended(String text, IStreamMonitor monitor)
	{
		String line;
		String resolvedOutputFilePath = resolveOutputFilePath(this.outputFile);
		try (BufferedReader bufferedReader = new BufferedReader(new StringReader(text));
				PrintWriter fileWriter = (resolvedOutputFilePath != null && !resolvedOutputFilePath.isEmpty())
						? new PrintWriter(new FileWriter(resolvedOutputFilePath, append), true)
						: new PrintWriter(Writer.nullWriter());)
		{
			while ((line = bufferedReader.readLine()) != null)
			{
				fileWriter.println(line);
```

bundles/com.espressif.idf.debug.gdbjtag.openocd/src/com/espressif/idf/debug/gdbjtag/openocd/dsf/process/StreamListener.java
```diff
@@ -50,11 +53,24 @@ public class StreamListener implements IStreamListener
...
+		this.outputFile = outputFile;
+		String resolvedOutputFile = resolveOutputFilePath(outputFile);
+		if (resolvedOutputFile != null && !resolvedOutputFile.isEmpty()) {
+			try {
+				fileWriter = new PrintWriter(new FileWriter(resolvedOutputFile, append), true);
+			} catch (IOException e) {
+				Logger.log(e);
+				fileWriter = null;
+			}
+		}
```
d-almazov: NIT: I think there's a way to make this a bit more optimized and easier to maintain. Since fileWriter is used only in one place and is autoclosable, we could avoid using it as a field and instead define the append flag and output file as fields.

bundles/com.espressif.idf.debug.gdbjtag.openocd/src/com/espressif/idf/debug/gdbjtag/openocd/dsf/process/StreamListener.java
```diff
@@ -186,5 +205,38 @@ public void dispose()
 		}
 		fErrorStreamMonitor = null;
 		fOutputStreamMonitor = null;
+		if (fileWriter != null) {
+			fileWriter.close();
+			fileWriter = null;
+		}
```
d-almazov: If we choose to use an autoclosable, we can avoid this.

sigmaaa:
Pushed a commit with minor improvements to the StreamListener class: made fileWriter autoclosable, avoided conditional checks by using nullWriter, and applied formatting.

AndriiFilippov:
@kolipakakondal 

Tested under:
OS - Windows 11 / Mac arm64 / Linux Ubuntu 24.04 
ESP-IDF: v5.4.2 
Boards: H2 / S3 

**Windows 11:**  everything works fine, I only have a few follow-up questions:

1. If I want to delete a file after the debug process has stopped, I can't do it because the pop-up says that the file is still occupied by an external process. There might be something wrong here, as I checked Task Manager for any active OpenOCD or GDB processes and I don't see any processes that could be preventing me from deleting the file after the debug process is finished. 
image
the logs for case when I run and stop debug process and trying to delete log file then:
[log.txt](https://github.com/user-attachments/files/21350105/log.txt)



2. I like the logic of case when user select the workspace to store log file, in this case the generated file has `.log` extension at the end, 
image
but in case of `File System...` selection the log file has just name and `Type: File`
image
also there is not option to select a type:
image
wouldn't it be more clear to add `.log` extension to any file selected, to make same structure for both cases? 

`Append` option works as expected ✅ 

**Mac arm64 / Linux Ubuntu 24.04** 

1.  The last line of the debug session output is **the only one** being written to the log file. 
the logs for case when open openocd.log file after debug is stopped:
[log_mac_arm64.txt](https://github.com/user-attachments/files/21350272/log_mac_arm64.txt)
Screenshot 2025-07-21 at 17 27 54

but with `Append` option selected - the openocd.log file contains full debug logs and everything works as expected ✅ 

The documentation is clear enough. ✅

AndriiFilippov:
@kolipakakondal  @alirana01 hi!
Please, review

kolipakakondal (commented):
Thanks @sigmaaa for improving this PR. LGTM

PR #1302, opened 2025-09-03
Dynamic ESP-IDF Target Discovery and Support Preview Targets

kolipakakondal:
## Description

This PR implements dynamic ESP-IDF target discovery that automatically reads both supported and preview targets from ESP-IDF's constants.py file, eliminating the need for manual plugin updates when new targets are added. The solution includes full backward compatibility and handles the toolchain unification

 **Key Features**
Dynamic targets: Automatically discovers all targets from ESP-IDF
Preview targets: Automatically discovers all preview targets
Future-proof: New targets appear without plugin updates going forward when esp-idf supports new targets
Zero maintenance: No manual plugin.xml updates needed
Backward Compatibility: Works with ESP-IDF v5.0+ (both target-specific and unified toolchain structures)

Fixes # ([IEP-XXX](https://jira.espressif.com:8443/browse/IEP-XXX))

## Type of change

## How has this been tested?

- Add esp-idf esp-idf version and verify that all the supported targets and preview targets are visible

**Test Configuration**:
* ESP-IDF Version:

- launch configuration
- build
- flash
- debugging
- 

## Checklist
- [ ] PR Self Reviewed
- [ ] Applied Code formatting
- [ ] Added Documentation
- [ ] Added Unit Test

## Summary by CodeRabbit

- New Features
  - Toolchains and targets are now discovered dynamically from your ESP-IDF installation; project target lists in the UI populate from those discoveries.

- Bug Fixes
  - Prevented a crash when determining the default board for targets that have no boards defined.

- Refactor
  - Removed bundled predefined toolchain entries; toolchain loading/activation now relies on ESP-IDF-derived target information.


bundles/com.espressif.idf.core/src/com/espressif/idf/core/toolchain/IDFTargets.java
```diff
@@ -0,0 +1,270 @@
...
+				return "xtensa"; //$NON-NLS-1$
+			case "esp32c2": //$NON-NLS-1$
+			case "esp32c3": //$NON-NLS-1$
+			case "esp32c6": //$NON-NLS-1$
+			case "esp32h2": //$NON-NLS-1$
+			case "esp32p4": //$NON-NLS-1$
+			default:
+				return "riscv32"; //$NON-NLS-1$
+			}
+		}
```
d-almazov, line 142: NIT:
We could simplify the getArchitecture() method to make it more readable and avoid multiple riscv32 statements or returns. For example, using Java 21 switch expressions:
```
public String getArchitecture() {
    return switch (name) {
        case "esp32", "esp32s2", "esp32s3" -> "xtensa";
        default -> "riscv32";
    };
}
```

Alternatively, a Set-based approach could be used
```
private static final Set<String> XTENSA_CHIPS = Set.of("esp32", "esp32s2", "esp32s3");

public String getArchitecture() {
    return XTENSA_CHIPS.contains(name) ? "xtensa" : "riscv32";
}
```

bundles/com.espressif.idf.core/src/com/espressif/idf/core/toolchain/IDFTargets.java
```diff
@@ -0,0 +1,270 @@
...
+				return "xtensa-" + name + "-elf"; //$NON-NLS-1$ //$NON-NLS-2$
+			case "esp32c2": //$NON-NLS-1$
+			case "esp32c3": //$NON-NLS-1$
+			case "esp32c6": //$NON-NLS-1$
+			case "esp32h2": //$NON-NLS-1$
+			case "esp32p4": //$NON-NLS-1$
+			default:
+				return "riscv32-esp-elf"; //$NON-NLS-1$
+			}
+		}
```
d-almazov, line 152: NIT: same as my earlier comment on getArchitecture() — we could simplify this in the same way using a switch expression or Set-based lookup.

sigmaaa (commented):
LGTM. Added a few nits: we could simplify the switch cases or use a Set for the xtensa targets. Additionally, it would be good to add unit tests for IDFTargetsReader.java to improve test coverage.

bundles/com.espressif.idf.core/src/com/espressif/idf/core/toolchain/enums/Target.java
```diff
@@ -0,0 +1,110 @@
...
+	{
+		return arch;
+	}
+
+	public static Target fromString(String s)
+	{
+		if (s == null)
+			return null;
+		String key = s.toLowerCase(Locale.ROOT);
+		for (Target t : values())
```
d-almazov: If a value isn’t present in the enum but comes from the IDF preview list, our current approach returns null and skips it. That doesn’t align with the PR’s goal

sigmaaa (commented):
The goal of this PR is to include targets from the preview list even if they aren’t explicitly defined in our target list, based on the assumption that **all new targets use the RICSV architecture**. Therefore, returning null when a target isn’t found in the enum isn’t the behavior we want.

bundles/com.espressif.idf.core/src/com/espressif/idf/core/toolchain/enums/Target.java
alirana01: @sigmaaa kindly take a look again the code is updated now

bundles/com.espressif.idf.core/src/com/espressif/idf/core/toolchain/IDFTargets.java
```diff
@@ -0,0 +1,182 @@
...
+
+	// ---------- Adders ----------
+
+	/** Strict for supported: unknown => throw (fail-fast). */
+	public void addSupportedTarget(String targetName)
+	{
+		Target known = Target.tryParse(targetName);
+		if (known == null)
+		{
+			throw new IllegalArgumentException("Unknown supported target: " + targetName); //$NON-NLS-1$
```
d-almazov: I don’t think this is the right approach. From my understanding, the goal of this PR is to avoid adding new targets directly in the code and instead rely fully on IDF, under the assumption that new targets will use the RISC-V architecture.

kolipakakondal (commented):
Thanks @alirana01 for the changes, LGTM

AndriiFilippov (approved):

Issue #1303, opened 2025-09-03
CreateProcess error=2 (IEP-1628)


kshitijgarg2609:
ESP-IDF works well on cmd, checked all installations and everything
I have installed this plugin
selected existing path of esp-idf
selected existing path of github
selected existing path of python
all tools gets downloaded, virtual python sets up well
after this plugin calls export.bat meanwhile setup
I have attached screenshot below

```
Executing D:\Installed\tmp\Espressif\Python\Scripts\python.exe D:\Installed\Espressif\esp-idf-v5.5.1\tools\idf_tools.py export --format=key-value
Executing null/Scripts/python.exe D:\Installed\Espressif\esp-idf-v5.5.1\tools\idf.py --version
Executing null/Scripts/python.exe D:\Installed\Espressif\esp-idf-v5.5.1\tools\idf.py --list-targets

```

<img width="819" height="416" alt="Image" src="https://github.com/user-attachments/assets/cd177cb6-1c8e-4a2c-bc90-612bd2d63a66" />
 

sharing all console logs here

```
Updating D:\Installed\tmp\Espressif\Tools\idf-env.json
Selected targets are: esp32c5, esp32c2, esp32s3, esp32c6, esp32h21, esp32h2, esp32s2, esp32p4, esp32c61, esp32h4, esp32c3, esp32
Current system platform: win64
Installing tools: xtensa-esp-elf-gdb, riscv32-esp-elf-gdb, xtensa-esp-elf, esp-clang, esp-clang-libs, riscv32-esp-elf, esp32ulp-elf, cmake, openocd-esp32, ninja, idf-exe, ccache, dfu-util, esp-rom-elfs, qemu-xtensa, qemu-riscv32
Skipping xtensa-esp-elf-gdb@16.2_20250324 (already installed)
Skipping riscv32-esp-elf-gdb@16.2_20250324 (already installed)
Skipping xtensa-esp-elf@esp-14.2.0_20241119 (already installed)
Skipping esp-clang@esp-19.1.2_20250312 (already installed)
Skipping esp-clang-libs@esp-19.1.2_20250312 (already installed)
Skipping riscv32-esp-elf@esp-14.2.0_20241119 (already installed)
Skipping esp32ulp-elf@2.38_20240113 (already installed)
Skipping cmake@3.30.2 (already installed)
Skipping openocd-esp32@v0.12.0-esp32-20250707 (already installed)
Skipping ninja@1.12.1 (already installed)
[...]

PR #1305, opened 2025-09-08
IEP-1628: Remove conflicting System env variables for export command

alirana01:
## Description
Original bug report: https://github.com/espressif/idf-eclipse-plugin/issues/1303

Some conflicting variables from the system environment can cause issues with the export script so we need to remove those.

Fixes # ([IEP-1628](https://jira.espressif.com:8443/browse/IEP-1628))

## Type of change

## How has this been tested?

Try to add IDF_PYTHON_ENV_PATH in your system env variables and run the IDE and try installing tools they should install without any problems

**Test Configuration**:
* ESP-IDF Version: any
* OS (Windows,Linux and macOS): Windows (POSIX based systems usually dont get this as long as they are not run from CLI adding the environment to the executable being run)

## Checklist
- [ ] PR Self Reviewed
- [ ] Applied Code formatting
- [ ] Verified on Windows

### DIFF
--- a/bundles/com.espressif.idf.ui/src/com/espressif/idf/ui/update/ExportIDFTools.java
+++ b/bundles/com.espressif.idf.ui/src/com/espressif/idf/ui/update/ExportIDFTools.java
@@ -8,6 +8,7 @@ import java.io.IOException;
 import java.text.MessageFormat;
 import java.util.ArrayList;
 import java.util.HashMap;
+import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
 
@@ -132,6 +133,8 @@ public class ExportIDFTools
 			addGitToEnvironment(environment, gitExePath);
 		}
 		
+		cleanUpSystemEnvironment(environment);
+		
 		final ProcessBuilderFactory processRunner = new ProcessBuilderFactory();
 		try
 		{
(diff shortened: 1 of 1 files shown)


alirana01 (commented):
Self review

bundles/com.espressif.idf.ui/src/com/espressif/idf/ui/update/ExportIDFTools.java
```diff
@@ -157,6 +160,26 @@ public IStatus getToolsExportOutputFromGivenIdfPath(final String pythonExePath,
...
+	/**
+	 * Remove the variables which can affect the idf_tools.py export command
+	 * These variables can come from system environment
+	 * @param environment
+	 */
+	private void cleanUpSystemEnvironment(Map<String, String> environment)
+	{
+		List<String> keysToRenmove = new LinkedList<>();
+		keysToRenmove.add(IDFEnvironmentVariables.IDF_PYTHON_ENV_PATH);
+		keysToRenmove.add(IDFEnvironmentVariables.IDF_PATH);
```
d-almazov: Don't we want to exclude other environment variables that have IDF in their name to avoid conflicts? For example, in IDFEnvironmentVariables we also have IDF_CCACHE_ENABLE, IDF_COMPONENT_MANAGER, IDF_MAINTAINER, and ESP_IDF_VERSION, which can potentially affect the plugin's functionality.

sigmaaa (commented):
LGTM! Could you also check the question I added in the review?

bundles/com.espressif.idf.ui/src/com/espressif/idf/ui/update/ExportIDFTools.java
alirana01: yeah that sounds about right I think I missed those
Thanks

alirana01:
@AndriiFilippov kindly verify after the build

bundles/com.espressif.idf.core/src/com/espressif/idf/core/util/IDFUtil.java
```diff
@@ -993,4 +994,29 @@ public static String runEspDetectConfigScript()
...
+		
+		for (String key : keysToRenmove)
+		{
+			if (environment.containsKey(key))
+			{
+				environment.remove(key);
+			}
+		}
+	}
 }
```
kolipakakondal, line 1022: Instead of cleaning up system environment variables, why can't we rely only on the environment variables configured in the IDE and that includes PATH as well? 

IDFEnvironmentVariables().getEnvMap()

Otherwise, if we go with the cleanup approach, we'll have to maintain this code all the time whenever there are changes in the IDF environment variables.

Please check?

kolipakakondal (changes requested):

PR #1309, opened 2025-09-15
added the version list for docs on master

alirana01:
Versions file for master added

## Summary by CodeRabbit

* **Documentation**
  * Introduced versioned documentation configuration to support a version selector.
  * Available versions: “latest” and “release-v4.0.0” (marked as pre-release).
  * Default hardware target set to ESP32; no multi-target content enabled yet.
  * Improves clarity on supported targets and upcoming release availability.


kolipakakondal:
I'm skeptical that this approach will work unless we move the docs. We will merge and see

PR #1409, opened 2026-03-07
IEP-1723: Improve New Project dialog initial startup time

sigmaaa:
## Description

Initially, I thought that the startup time of the New Project dialog was affected by template loading. However, after investigating, I found that the delay was caused by the check for space support.
To perform this check, we had to run an esp-idf script to verify whether the esp-idf version was 5.0 or higher. Since we now support only esp-idf 5.x, this check is no longer necessary.
I removed the related code, and the startup delay is now gone.

Fixes # ([IEP-1723](https://jira.espressif.com:8443/browse/IEP-1723))

## Type of change

## How has this been tested?

Create a new ESP-IDF project. The dialog should open without any waiting time.

**Test Configuration**:
* ESP-IDF Version:

- New ESP-IDF project wizard

## Checklist
- [ ] PR Self Reviewed
- [ ] Applied Code formatting
- [ ] Added Documentation
- [ ] Added Unit Test

## Summary by CodeRabbit

* **Bug Fixes**
  * Project creation wizard no longer blocks or errors on spaces in project names or workspace locations.
  * Build preflight no longer blocks builds due to spaces in project paths; builds proceed when other checks pass.

* **Tests**
  * Removed unit tests that validated space-related behavior and adjusted test visibility accordingly.

### DIFF
--- a/tests/com.espressif.idf.core.test/src/com/espressif/idf/core/util/test/IDFUtilTest.java
+++ b/tests/com.espressif.idf.core.test/src/com/espressif/idf/core/util/test/IDFUtilTest.java
@@ -1,9 +1,7 @@
 package com.espressif.idf.core.util.test;
 
 import static org.junit.jupiter.api.Assertions.assertEquals;
-import static org.junit.jupiter.api.Assertions.assertFalse;
 import static org.junit.jupiter.api.Assertions.assertNull;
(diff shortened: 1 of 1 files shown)


AndriiFilippov:
@sigmaaa hi ! 

LGTM 👍

